PR #84, opened 2026-03-20
Bypass serialization on self-dispatch while retaining traceability and context propagation — Closes #80

### LINKED ISSUE #80: Bypass serialization on self-dispatch while retaining traceability and context propagation
## Description

When a worker dispatches a routine that routes back to itself (self-dispatch), the current path serializes the task payload with cloudpickle, wraps it in protobuf, sends it over a gRPC bidirectional stream to localhost, deserializes it on the receiving end, and executes it. The cloudpickle round-trip is unnecessary when the dispatcher and executor are in the same process — the callable, args, kwargs, and proxy are already live Python objects that can be passed directly.

Add a `PassthroughSerializer` that stores the four payload fields (callable, args, kwargs, proxy) in a process-local dict keyed by a nonce, and writes only the nonce into the protobuf bytes fields. The request still travels through gRPC so the full streaming protocol, traceability, cancellation, and context propagation are preserved — only the serialization cost is eliminated.

## Motivation

Self-dispatch currently cloudpickles every payload field on the send side and unpickles on the receive side — roughly 0.5-2ms per hop depending on payload complexity. For nested routines (routine calling routine), this overhead compounds at every nesting level. In a single-worker pool, every nested dispatch is guaranteed to self-dispatch. Bypassing serialization eliminates this cost while keeping the gRPC dispatch path and all its invariants intact.

## Expected outcome

Self-dispatched tasks skip cloudpickle serialization/deserialization of callable, args, kwargs, and proxy fields. The gRPC streaming protocol, task tags, UUIDs, logging, RuntimeContext propagation, cancellation, and error propagation behave identically to remote dispatch.

conradbzura:
## Summary

When a worker dispatches a routine that routes back to itself, the current path cloudpickles every payload field even though both sides are in the same process. Introduce a `Serializer` protocol and a `PassthroughSerializer` implementation that stores objects in a process-local `WeakKeyDictionary` keyed by UUID sentinels, returning only the raw UUID bytes to the protobuf layer. The gRPC streaming path is unchanged — only the serialization cost is eliminated for self-dispatch.

The `PassthroughSerializer` instance doubles as a scope guard: it holds strong references to the store keys, and when it goes out of scope the `WeakKeyDictionary` entries are garbage-collected automatically.

Closes #80

### Proto schema extension

Add an optional `serializer` field to the Task message. When absent, the receiver falls back to cloudpickle with zero overhead for existing remote dispatch. When present, the receiver recovers the serializer and uses it for payload deserialization.

### Serializer abstraction

Define a `Serializer` runtime-checkable protocol (`dumps`/`loads`) that `cloudpickle` already satisfies. Implement `PassthroughSerializer` backed by a module-level `WeakKeyDictionary` — `dumps` stores the object and returns a 16-byte UUID token, `loads` pops the entry by reconstructing the key from the token. All instances hash and compare equal so LRU caching works across pickle boundaries.

Extend `Task.to_protobuf` and `Task.from_protobuf` to accept and detect the serializer. The proxy field is always cloudpickled unless the serializer is a `PassthroughSerializer` — custom serializers are not trusted for proxy serialization.

### Self-dispatch detection

`WorkerConnection.dispatch()` compares the target address against the current process's `WorkerMetadata`. On match, it creates a `PassthroughSerializer` and passes it to `to_protobuf`, bypassing cloudpickle for callable, args, kwargs, and proxy. The serializer is scoped to the dispatch lifetime for automatic cleanup.

### Worker identity context

Add `__worker_metadata__` and `__worker_service__` ContextVars to the package root, set by the worker subprocess during server startup. Refactor `WorkerProcess` to send full `WorkerMetadata` (including uid, tags, extra) back to the parent via pipe, replacing the previous port-only pipe.

### Test cases

| 1 | `TestTask` | A Task instance | `to_protobuf()` called with no serializer | Serializer field is not set | Default serialization |
| 2 | `TestTask` | A Task and a PassthroughSerializer | `to_protobuf(serializer=...)` called | Serializer field is set | Serializer field presence |
| 3 | `TestTask` | A protobuf Task with no serializer field | `from_protobuf()` called | All fields deserialized via cloudpickle | Backward compatibility |
| 4 | `TestTask` | A Task round-tripped via PassthroughSerializer | `from_protobuf()` called | All public attributes preserved by identity | Passthrough roundtrip |
| 5 | `TestPassthroughSerializer` | An arbitrary Python object | `dumps` called | Returns a 16-byte token | Token format |
| 6 | `TestPassthroughSerializer` | Two distinct objects | `dumps` called on each | Returns different tokens | Token uniqueness |
| 7 | `TestPassthroughSerializer` | An object stored via `dumps` | `loads` called with the token | Returns the exact same object by identity | Roundtrip identity |
| 8 | `TestPassthroughSerializer` | An object stored via `dumps` | `loads` called twice with the same token | Raises KeyError on second call | Entry consumption |
| 9 | `TestPassthroughSerializer` | A PassthroughSerializer instance | Checked via isinstance against Serializer | Recognized as a Serializer | Protocol conformance |
| 10 | `TestPassthroughSerializer` | A PassthroughSerializer instance | Cloudpickled and unpickled | Produces a functional instance | Pickle survivability |
| 11 | `TestWorkerConnection` | Worker metadata address matches target | `dispatch(task)` called | Serializer field set, payload not cloudpickled | Self-dispatch serializer |
| 12 | `TestWorkerConnection` | Worker metadata address differs from target | `dispatch(task)` called | Serializer field not set | Address mismatch fallback |
| 13 | `TestWorkerConnection` | No worker metadata set | `dispatch(task)` called | Serializer field not set | Non-worker fallback |

### DIFF
--- a/wool/src/wool/runtime/routine/task.py
+++ b/wool/src/wool/runtime/routine/task.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import asyncio
+import functools
 import logging
+import weakref
 import traceback
 from collections.abc import Callable
 from contextlib import contextmanager
@@ -10,6 +12,7 @@ from dataclasses import dataclass
 from inspect import isasyncgenfunction
 from inspect import iscoroutinefunction
 from types import TracebackType
+from typing import Any
 from typing import AsyncGenerator
 from typing import ContextManager
 from typing import Coroutine
@@ -24,6 +27,7 @@ from typing import cast
 from typing import overload
 from typing import runtime_checkable
 from uuid import UUID
+from uuid import uuid4
 
 import cloudpickle
 
@@ -36,6 +40,87 @@ Kwargs = Dict
 Timeout = SupportsInt
 Routine: TypeAlias = Coroutine | AsyncGenerator
 W = TypeVar("W", bound=Routine)
+
+
+# public
+@runtime_checkable
+class Serializer(Protocol):
+    """Protocol for pluggable serialization of Task payload fields."""
+
+    def dumps(self, obj: Any) -> bytes: ...
+
+    def loads(self, data: bytes) -> Any: ...
+
+
+class _PassthroughKey:
+    """Weak-referenceable key for the passthrough store."""
+
+    __slots__ = ("__weakref__", "token")
+
+    def __init__(self, token: UUID | None = None):
+        self.token = token if token is not None else uuid4()
+
+    def __hash__(self):
+        return hash(self.token)
+
+    def __eq__(self, other):
+        return isinstance(other, _PassthroughKey) and self.token == other.token
+
+
+class PassthroughSerializer:
+    """In-process serializer that avoids pickling entirely.
+
+    Each instance acts as a scope guard for one dispatch.
+    ``dumps`` creates a weakly-referenceable key, stores the object
+    in a module-level :class:`~weakref.WeakKeyDictionary`, and
+    retains a strong reference to the key on ``self``.  When the
+    serializer goes out of scope the keys are garbage-collected and
+    the weak-dict entries are removed automatically.
+
+    ``loads`` is static — it reconstructs the key from the bytes
+    token in the protobuf message and pops the entry from the store.
+
+    All instances hash and compare equal so that
+    ``_pickle_serializer``'s LRU cache hits on every call.
+    """
+
+    def __init__(self):
+        self._keys: list[_PassthroughKey] = []
+
+    def __hash__(self):
+        return hash(PassthroughSerializer)
+
+    def __eq__(self, other):
+        return isinstance(other, PassthroughSerializer)
+
+    def __reduce__(self):
+        return (PassthroughSerializer, ())
+
+    def dumps(self, obj: Any) -> bytes:
+        key = _PassthroughKey()
+        self._keys.append(key)
+        _passthrough_store[key] = obj
+        return key.token.bytes
+
+    @staticmethod
+    def loads(data: bytes) -> Any:
+        key = _PassthroughKey(UUID(bytes=data))
+        return _passthrough_store.pop(key)
+
+
+_passthrough_store: weakref.WeakKeyDictionary[_PassthroughKey, Any] = (
+    weakref.WeakKeyDictionary()
+)
+
+
+@functools.lru_cache(maxsize=8)
+def _pickle_serializer(s: Serializer) -> bytes:
+    return cloudpickle.dumps(s)
+
+
+@functools.lru_cache(maxsize=8)
+def _unpickle_serializer(data: bytes) -> Serializer:
+    return cloudpickle.loads(data)
 
 
 _do_dispatch: ContextVar[bool] = ContextVar("_do_dispatch", default=True)
@@ -209,6 +294,9 @@ class Task(Generic[W]):
     def from_protobuf(cls, task: protocol.Task) -> Task:
         """Deserialize a Task from a protobuf message.
 
+        .. note::
+            The payload's serializer is unpickled and cached for subsequent calls.
+
         :param task:
             A protobuf ``Task`` message.
         :returns:
@@ -219,32 +307,62 @@ class Task(Generic[W]):
             if task.HasField("context")
             else None
         )
+        if task.HasField("serializer"):
+            s = _unpickle_serializer(task.serializer)
+            loads = s.loads
+            if isinstance(s, PassthroughSerializer):
+                proxy_loads = s.loads
+            else:
+                proxy_loads = cloudpickle.loads
+        else:
+            loads = cloudpickle.loads
+            proxy_loads = cloudpickle.loads
         return cls(
             id=UUID(task.id),
-            callable=cloudpickle.loads(task.callable),
-            args=cloudpickle.loads(task.args),
-            kwargs=cloudpickle.loads(task.kwargs),
+            callable=loads(task.callable),
+            args=loads(task.args),
+            kwargs=loads(task.kwargs),
             caller=UUID(task.caller) if task.caller else None,
-            proxy=cloudpickle.loads(task.proxy),
+            proxy=proxy_loads(task.proxy),
             timeout=task.timeout if task.timeout else 0,
             tag=task.tag if task.tag else None,
             context=context,
         )
 
-    def to_protobuf(self) -> protocol.Task:
-        return protocol.Task(
+    def to_protobuf(self, serializer: Serializer | None = None) -> protocol.Task:
+        """Serialize this Task to a protobuf message.
+
+        :param serializer:
+            Optional serializer for the callable and its arguments. When ``None`` (the
+            default), ``cloudpickle`` is used and the protobuf ``serializer`` field is
+            left unset. When provided, the serializer is pickled into the ``serializer``
+            field so that :meth:`from_protobuf` can use it on the receiving side.
+
+            .. note::
+                If specified, the serializer is pickled and cached for subsequent calls.
+        :returns:
+            A protobuf ``Task`` message.
+        """
+        dumps = serializer.dumps if serializer is not None else cloudpickle.dumps
+        proxy_dumps = (
+            dumps if isinstance(serializer, PassthroughSerializer) else cloudpickle.dumps
+        )
+        task_msg = protocol.Task(
             version=protocol.__version__,
             id=str(self.id),
-            callable=cloudpickle.dumps(self.callable),
-            args=cloudpickle.dumps(self.args),
-            kwargs=cloudpickle.dumps(self.kwargs),
+            callable=dumps(self.callable),
+            args=dumps(self.args),
+            kwargs=dumps(self.kwargs),
             caller=str(self.caller) if self.caller else "",
-            proxy=cloudpickle.dumps(self.proxy),
+            proxy=proxy_dumps(self.proxy),
             proxy_id=str(self.proxy.id),
             timeout=int(self.timeout) if self.timeout else 0,
             tag=self.tag if self.tag else "",
             context=self.context.to_protobuf() if self.context else None,
         )
+        if serializer is not None:
+            task_msg.serializer = _pickle_serializer(serializer)
+        return task_msg
 
     def dispatch(self) -> W:
         if isasyncgenfunction(self.callable):
--- a/wool/src/wool/runtime/routine/task.py
+++ b/wool/src/wool/runtime/routine/task.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import asyncio
 import functools
 import logging
-import weakref
 import traceback
+import weakref
 from collections.abc import Callable
 from contextlib import contextmanager
 from contextvars import ContextVar
@@ -64,7 +64,9 @@ class _PassthroughKey:
         return hash(self.token)
 
     def __eq__(self, other):
-        return isinstance(other, _PassthroughKey) and self.token == other.token
+        if not isinstance(other, _PassthroughKey):
+            return super().__eq__(other)
+        return self.token == other.token
 
 
 class PassthroughSerializer:
@@ -91,6 +93,10 @@ class PassthroughSerializer:
         return hash(PassthroughSerializer)
 
     def __eq__(self, other):
+        # Required by _pickle_serializer's lru_cache: all instances
+        # share the same hash, so __eq__ must confirm the match for
+        # the cache to hit instead of treating each instance as a
+        # separate key.
         return isinstance(other, PassthroughSerializer)
 
     def __reduce__(self):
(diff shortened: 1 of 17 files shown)


wool/src/wool/runtime/worker/connection.py
```diff
@@ -388,10 +390,24 @@ async def dispatch(
         if timeout is not None and timeout <= 0:
             raise ValueError("Dispatch timeout must be positive")
 
+        # Self-dispatch optimisation: when dispatching to the current
+        # worker process, use a PassthroughSerializer so the four
+        # payload fields (callable, args, kwargs, proxy) are stored
+        # in-process instead of being cloudpickled.  The request
+        # still travels through gRPC so the full streaming protocol
+        # is preserved.
+        metadata = wool.__worker_metadata__.get()
```
conradbzura: Set this in the if-statement using the walrus operator.

wool/src/wool/runtime/routine/task.py
```diff
@@ -219,32 +304,54 @@ def from_protobuf(cls, task: protocol.Task) -> Task:
...
             id=str(self.id),
-            callable=cloudpickle.dumps(self.callable),
-            args=cloudpickle.dumps(self.args),
-            kwargs=cloudpickle.dumps(self.kwargs),
+            callable=dumps(self.callable),
+            args=dumps(self.args),
+            kwargs=dumps(self.kwargs),
             caller=str(self.caller) if self.caller else "",
-            proxy=cloudpickle.dumps(self.proxy),
+            proxy=dumps(self.proxy),
```
conradbzura: We need to verify that we don't attempt to re-enter this on the worker side when the task is executed.

I also don't think we want to serialize this with user-defined serializers - we need to retain control over this behavior. We should ONLY use the passed serializer here if it's a PassthroughSerializer.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -388,10 +390,24 @@ async def dispatch(
...
+        if metadata is not None and metadata.address == self._target:
+            serializer = PassthroughSerializer()
+            pb_task = task.to_protobuf(serializer=serializer)
```
conradbzura: To avoid the temptation to simplify this to `task.to_protobuf(serializer=PassthroughSerializer())` and lose the reference immediately, add a comment and move this expression to line 410, i.e.:
```python
call = await self._dispatch(ch, task.to_protobuf(serializer=serializer), timeout)
```

wool/src/wool/runtime/worker/connection.py
```diff
@@ -388,10 +390,24 @@ async def dispatch(
...
+        else:
+            serializer = None
+            pb_task = task.to_protobuf()
```
conradbzura: This can be dropped once the previous comment is resolved.

wool/src/wool/runtime/routine/task.py
```diff
@@ -219,32 +304,54 @@ def from_protobuf(cls, task: protocol.Task) -> Task:
...
+        """Serialize this Task to a protobuf message.
+
+        :param serializer:
+            Optional serializer for the four payload fields
+            (callable, args, kwargs, proxy).  When ``None``
+            (the default), ``cloudpickle`` is used and the
+            protobuf ``serializer`` field is left unset.
+            When provided, the serializer is pickled into the
+            ``serializer`` field so that :meth:`from_protobuf`
+            can auto-detect and use it on the receiving side.
```
conradbzura: We should describe the caching mechanism here so that users are aware.

wool/src/wool/runtime/routine/task.py
```diff
@@ -38,6 +42,87 @@
...
+    __slots__ = ("__weakref__", "token")
+
+    def __init__(self, token: UUID | None = None):
+        self.token = token if token is not None else uuid4()
+
+    def __hash__(self):
+        return hash(self.token)
+
+    def __eq__(self, other):
+        return isinstance(other, _PassthroughKey) and self.token == other.token
```
conradbzura: If other is not a _PassthroughKey, call `super().__eq__(...)`.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -388,10 +390,24 @@ async def dispatch(
         if timeout is not None and timeout <= 0:
             raise ValueError("Dispatch timeout must be positive")
 
+        # Self-dispatch optimisation: when dispatching to the current
+        # worker process, use a PassthroughSerializer so the four
+        # payload fields (callable, args, kwargs, proxy) are stored
+        # in-process instead of being cloudpickled.  The request
+        # still travels through gRPC so the full streaming protocol
+        # is preserved.
```
conradbzura: This comment should be a note admonition in the docstring.

wool/src/wool/runtime/worker/process.py
```diff
@@ -219,6 +222,19 @@ async def _serve(self):
...
+                    # Publish worker identity for self-dispatch detection
+                    wool.__worker_metadata__.set(
+                        WorkerMetadata(
+                            uid=uuid.uuid4(),
+                            address=self._address(self._host, port),
+                            pid=os.getpid(),
+                            version=protocol.__version__,
+                            secure=self._credentials is not None,
+                        )
+                    )
```
conradbzura, line 264: This duplicates the logic in LocalWorker, it would be good to either refactor this into a helper function or have the process create it and send it back to the parent process via a pipe (I think I prefer this alternative - it's already serializable to protobuf). This should replace the port pipe. The process would also need `tags` and `extra` from the parent.

wool/src/wool/runtime/routine/task.py
```diff
@@ -219,32 +304,54 @@ def from_protobuf(cls, task: protocol.Task) -> Task:
...
+            (the default), ``cloudpickle`` is used and the
+            protobuf ``serializer`` field is left unset.
+            When provided, the serializer is pickled into the
+            ``serializer`` field so that :meth:`from_protobuf`
+            can auto-detect and use it on the receiving side.
+        :returns:
+            A protobuf ``Task`` message.
+        """
+        dumps = serializer.dumps if serializer is not None else cloudpickle.dumps
+        pb = protocol.Task(
```
conradbzura: Call this `task_msg` instead of `pb`.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -421,13 +437,13 @@ async def close(self):
         except KeyError:
             pass
 
-    async def _dispatch(self, ch, task, timeout):
+    async def _dispatch(self, ch, pb_task, timeout):
```
conradbzura: Call this `task_msg` instead of `pb_task` and add type hints to the signature. Rename `ch` to `channel`.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+        task = Task(
+            id=uuid4(),
+            callable=sample_async_callable,
+            args=(),
+            kwargs={},
+            proxy=picklable_proxy,
+        )
+
+        # Act
+        pb_task = task.to_protobuf()
```
conradbzura: Use the convention `*_msg` instead of `pb_*` throughout.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+
+    @settings(max_examples=50, deadline=None)
+    @given(
+        task_id=st.uuids(),
+        timeout=st.integers(min_value=0, max_value=3600),
+        caller_id=st.one_of(st.none(), st.uuids()),
+        tag=st.one_of(st.none(), st.text(min_size=1, max_size=100)),
+    )
+    @pytest.mark.asyncio
+    async def test_to_protobuf_with_passthrough_serializer(
```
conradbzura: Since this test is mainly evaluating the result of `from_protobuf`, rename this test `test_from_protobuf_*`.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+        caller_id,
+        tag,
+    ):
+        """Test Task protobuf roundtrip via PassthroughSerializer.
+
+        Given:
+            A Task with valid data and a PassthroughSerializer.
+        When:
+            ``to_protobuf(serializer=...)`` then ``from_protobuf()``
+            is called.
```
conradbzura: `to_protobuf(serializer=...)` arranges the test and should be part of the "given".

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+            kwargs=kwargs,
+            proxy=proxy,
+            timeout=timeout,
+            caller=caller_id,
+            tag=tag,
+        )
+        serializer = PassthroughSerializer()
+
+        # Act
+        pb_task = original_task.to_protobuf(serializer=serializer)
```
conradbzura: This should be part of the "arrange" step.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+        assert deserialized_task.callable is original_task.callable
+        assert deserialized_task.args is original_task.args
+        assert deserialized_task.kwargs is original_task.kwargs
+        assert deserialized_task.caller == original_task.caller
+        assert deserialized_task.proxy is original_task.proxy
+        assert deserialized_task.timeout == original_task.timeout
+        assert deserialized_task.tag == original_task.tag
+
+
+# --- TestPassthroughSerializer ---
```
conradbzura: Don't add structural comments.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+
+
+class TestPassthroughSerializer:
+    """Tests for :py:class:`PassthroughSerializer`."""
+
+    def test_dumps_loads_roundtrip_with_object(self):
```
conradbzura: Again, treat `dumps` as the precondition (i.e., part of the "given") and `loads` as the "when" - the name should reflect this – `test_loads_*`.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+        assert deserialized_task.caller == original_task.caller
+        assert deserialized_task.proxy is original_task.proxy
+        assert deserialized_task.timeout == original_task.timeout
+        assert deserialized_task.tag == original_task.tag
+
+
+# --- TestPassthroughSerializer ---
+
+
+class TestPassthroughSerializer:
```
conradbzura, line 1624: We need test cases specifically for `dumps()`.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+        restored = cloudpickle.loads(cloudpickle.dumps(original))
+        obj = {"test": True}
+        data = restored.dumps(obj)
+        result = restored.loads(data)
+
+        # Assert
+        assert result is obj
+
 
 # --- TestTaskException ---
```
conradbzura: Let's clean up all of the structural comments in this file.

wool/tests/runtime/worker/test_connection.py
```diff
@@ -1168,3 +1169,154 @@ async def test_dispatch_with_custom_options(
             "grpc.max_send_message_length",
             50 * 1024 * 1024,
         ) in call_options
+
+    # ------------------------------------------------------------------ #
+    # Self-dispatch serializer optimisation tests
+    # ------------------------------------------------------------------ #
```
conradbzura: Don't add structural comments.

wool/tests/runtime/routine/test_task.py
```diff
@@ -1472,6 +1474,254 @@ async def test_callable():
...
+    ):
+        """Test to_protobuf without serializer omits the serializer field.
+
+        Given:
+            A Task instance.
+        When:
+            ``to_protobuf()`` is called with no serializer argument.
+        Then:
+            It should not set the ``serializer`` field on the
+            protobuf message.
```
conradbzura: The given-when-then descriptions don't require punctation – in fact, it should be omitted since they should be read together as parts of a single sentence.

wool/src/wool/runtime/worker/process.py
```diff
@@ -94,15 +103,27 @@ def __init__(
...
+        given, or ``None`` when port is 0 (random).
+
         :returns:
-            The address in "host:port" format, or None if not started.
+            The address in "host:port" format, or None if not started
+            and port is 0.
         """
+        if self._metadata is not None:
+            return self._metadata.address
         return self._address(self._host, self._port) if self._port else None
```
conradbzura: This line can be dropped.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -388,10 +398,22 @@ async def dispatch(
...
+        else:
+            serializer = None
+
         ch = await _channel_pool.acquire(self._key)
         try:
             try:
-                call = await self._dispatch(ch, task, timeout)
+                # The serializer must remain in scope for the
+                # lifetime of the gRPC call so its strong refs to
+                # the PassthroughSerializer keys are not collected.
```
conradbzura: On a second thought, drop this comment.

wool/src/wool/runtime/routine/task.py
```diff
@@ -38,6 +42,89 @@
...
+    """
+
+    def __init__(self):
+        self._keys: list[_PassthroughKey] = []
+
+    def __hash__(self):
+        return hash(PassthroughSerializer)
+
+    def __eq__(self, other):
+        return isinstance(other, PassthroughSerializer)
```
conradbzura, line 100: Do we have to define an equality comparator? Is it used anywhere?

PR #98, opened 2026-03-22
Use Unix domain sockets for self-dispatch gRPC connections — Closes #96

### LINKED ISSUE #96: Use Unix domain sockets for self-dispatch gRPC connections
## Description

When `WorkerConnection.dispatch()` detects self-dispatch (target address matches the current worker's metadata), create the gRPC channel over a Unix domain socket instead of a TCP loopback connection. The worker subprocess binds a UDS path in addition to its TCP port during server startup, and self-dispatch connections use the `unix:` target scheme to avoid the TCP stack entirely. On platforms where UDS is unavailable, fall back to the existing TCP loopback path.

## Motivation

Self-dispatch currently routes through TCP loopback even though both endpoints are in the same process. While the `PassthroughSerializer` (#80) eliminates serialization overhead, the connection still traverses the kernel's TCP stack — SYN/ACK handshake, Nagle's algorithm, socket buffer copies, and per-connection FD overhead for both the listening socket and the accepted connection. UDS eliminates the TCP protocol machinery and reduces the kernel path to a single buffer copy between socket buffers, cutting per-dispatch latency and FD pressure for nested routines.

## Expected outcome

Self-dispatched tasks connect over a Unix domain socket bound by the worker subprocess on platforms that support it. On platforms without UDS support, the existing TCP loopback path is used. Remote dispatch is unaffected — the TCP port remains the primary transport for cross-process and cross-host connections. The UDS path is cleaned up when the worker process exits.

conradbzura:
## Summary

Bind a Unix domain socket alongside the TCP port during worker subprocess startup and route self-dispatch gRPC connections through it to bypass the TCP stack. On platforms without `AF_UNIX` support, fall back to the existing TCP loopback path. Remote dispatch is unaffected. Clean up the UDS socket file and channel pool entry when the worker process exits or the connection is closed.

The UDS address is stored as a plain module-level variable (`wool.__worker_uds_address__`) rather than on `WorkerMetadata`, since it is a local transport detail with no meaning for remote workers or discovery protocols. As part of this change, simplify `wool.__worker_metadata__` from a `ContextVar` to a plain module variable, since both are set once per worker process and never vary per task.

Additionally, extract the integration test xfail helper into a data-driven `_KnownBug` registry with an `xfail_known_bugs` fixture that retries `StatusCode.INTERNAL` errors with exponential backoff before falling through to xfail. Update the test guide to document the new convention.

Closes #96

### Bind UDS path in worker subprocess

Add a `_HAS_UDS` platform guard (`hasattr(socket, "AF_UNIX")`) and, when available, bind an additional insecure port at `unix:{tempdir}/wool-{uid}.sock` during `WorkerProcess._serve()`. Clean up any stale socket file before binding and unlink the path in the `finally` block on server shutdown.

### Route self-dispatch through UDS

In `WorkerConnection.dispatch()`, when self-dispatch is detected and `wool.__worker_uds_address__` is set, construct a channel pool key using the UDS target with `None` credentials (UDS is kernel-secured, no TLS needed). Thread the effective key through `_execute()` so both acquire and release use the same pool entry. Clear both TCP and UDS pool entries in `close()`.

### Simplify worker identity state to plain variables

Replace `wool.__worker_metadata__` (a `ContextVar`) with a plain module-level variable and add `wool.__worker_uds_address__` as another plain variable. Both are set once per worker process in `_serve()` and read during dispatch — per-task isolation is unnecessary.

### Extract xfail_known_bugs fixture with retry

Replace the inline `_xfail_known_bugs` helper with a data-driven `_KnownBug` registry and an `xfail_known_bugs` fixture. Each bug entry defines a `retryable` predicate and retry count — `StatusCode.INTERNAL` errors from the grpc race are retried with exponential backoff before falling through to xfail. Non-retryable matches xfail immediately. Unmatched exceptions propagate normally. Update the test guide to document the new xfail convention.

## Test cases

|---|---|---|---|---|---|
| 1 | `TestWorkerConnection` | WorkerConnection target matches worker metadata and UDS address is set | `dispatch()` is called | Channel is created with the UDS address, None credentials, and PassthroughSerializer | UDS key construction in dispatch |
| 2 | `TestWorkerConnection` | WorkerConnection and UDS address is available | `close()` is called | Both TCP and UDS pool entries are cleared without error | UDS pool cleanup in close |
| 3 | `TestWorkerConnection` | WorkerConnection target matches worker metadata and no UDS address | `dispatch()` is called | PassthroughSerializer is used with TCP key | Self-dispatch TCP fallback |
| 4 | `TestWorkerConnection` | WorkerConnection target differs from worker metadata | `dispatch()` is called | Cloudpickle is used with TCP key | Non-self-dispatch path |
| 5 | `TestWorkerConnection` | No worker metadata set | `dispatch()` is called | Cloudpickle is used with TCP key | No-metadata path |
| 6 | `TestWorkerProcess` | WorkerProcess with UDS disabled | Server is configured | Only one TCP port is bound | Single TCP port invariant |
| 7 | `TestWorkerProcess` | Secure WorkerProcess with UDS disabled | Server is configured | No insecure TCP port is bound | No insecure TCP backdoor |

### DIFF
--- a/wool/src/wool/runtime/worker/connection.py
+++ b/wool/src/wool/runtime/worker/connection.py
@@ -20,6 +20,7 @@ from wool.runtime.routine.task import Task
 from wool.runtime.worker.base import WorkerOptions
 
 _DispatchCall: TypeAlias = grpc.aio.StreamStreamCall[protocol.Request, protocol.Response]
+_PoolKey: TypeAlias = tuple[str, grpc.ChannelCredentials | None, int, WorkerOptions]
 
 _T = TypeVar("_T")
 
@@ -353,6 +354,7 @@ class WorkerConnection:
         self._limit = limit
         self._options = options if options is not None else WorkerOptions()
         self._key = (target, credentials, limit, self._options)
+        self._uds_key: _PoolKey | None = None
 
     async def dispatch(
         self,
@@ -399,17 +401,23 @@ class WorkerConnection:
             raise ValueError("Dispatch timeout must be positive")
 
         if (
-            metadata := wool.__worker_metadata__.get()
+            metadata := wool.__worker_metadata__
         ) is not None and metadata.address == self._target:
             serializer = PassthroughSerializer()
+            if (uds_address := wool.__worker_uds_address__) is not None:
+                key = (uds_address, None, self._limit, self._options)
+                self._uds_key = key
+            else:
+                key = self._key
         else:
             serializer = None
+            key = self._key
 
-        ch = await _channel_pool.acquire(self._key)
+        channel = await _channel_pool.acquire(key)
         try:
             try:
                 call = await self._dispatch(
-                    ch, task.to_protobuf(serializer=serializer), timeout
+                    channel, task.to_protobuf(serializer=serializer), timeout
                 )
             except grpc.RpcError as error:
                 code = error.code()
@@ -419,26 +427,31 @@ class WorkerConnection:
                 else:
                     raise RpcError(code, details) from error
 
-            gen = self._execute(call, task)
-            await gen.__anext__()  # Prime: _execute acquires its own ref
+            stream = self._execute(call, task, key)
+            await stream.__anext__()  # Prime: _execute acquires its own ref
         except BaseException:
-            await _channel_pool.release(self._key)
+            await _channel_pool.release(key)
             raise
 
-        await _channel_pool.release(self._key)
-        return cast(AsyncGenerator[protocol.Message, None], gen)
+        await _channel_pool.release(key)
+        return cast(AsyncGenerator[protocol.Message, None], stream)
 
     async def close(self):
         """Close the connection and release all pooled resources.
 
-        Clears the pooled channel entry for this connection's key.
-        Idempotent: safe to call multiple times or on connections
-        that were never used.
+        Clears the pooled channel entries for both the TCP key and,
+        if a UDS address is available, the UDS key. Idempotent: safe
+        to call multiple times or on connections that were never used.
         """
         try:
             await _channel_pool.clear(self._key)
         except KeyError:
             pass
+        if self._uds_key is not None:
+            try:
+                await _channel_pool.clear(self._uds_key)
+            except KeyError:
+                pass
 
     async def _dispatch(
         self,
@@ -475,9 +488,9 @@ class WorkerConnection:
         return call
 
     async def _execute(
-        self, call: _DispatchCall, task: Task
+        self, call: _DispatchCall, task: Task, key: _PoolKey
     ) -> AsyncGenerator[protocol.Message | None, None]:
-        ch = await _channel_pool.acquire(self._key)
+        channel = await _channel_pool.acquire(key)
         try:
             yield  # Priming yield — signals dispatch() that ref is held
             stream = _DispatchStream(call, task)
@@ -503,6 +516,6 @@ class WorkerConnection:
                     pass
                 raise
             finally:
-                ch.semaphore.release()
+                channel.semaphore.release()
         finally:
-            await _channel_pool.release(self._key)
+            await _channel_pool.release(key)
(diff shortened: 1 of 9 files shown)


wool/src/wool/runtime/worker/connection.py
```diff
@@ -399,13 +399,19 @@ async def dispatch(
             raise ValueError("Dispatch timeout must be positive")
 
         if (
-            metadata := wool.__worker_metadata__.get()
+            metadata := wool.__worker_metadata__
         ) is not None and metadata.address == self._target:
             serializer = PassthroughSerializer()
+            uds_address = wool.__worker_uds_address__
+            if uds_address is not None:
```
conradbzura: Replace with:

```python
if (uds_address := wool.__worker_uds_address__) is not None:
```

wool/src/wool/runtime/worker/connection.py
```diff
@@ -399,13 +399,19 @@ async def dispatch(
...
+                key = (uds_address, None, self._limit, self._options)
+            else:
+                key = self._key
         else:
             serializer = None
+            key = self._key
 
-        ch = await _channel_pool.acquire(self._key)
+        ch = await _channel_pool.acquire(key)
```
conradbzura: Rename `ch` to `channel` (do this throughout).

wool/src/wool/runtime/worker/connection.py
```diff
@@ -419,13 +425,13 @@ async def dispatch(
                 else:
                     raise RpcError(code, details) from error
 
-            gen = self._execute(call, task)
+            gen = self._execute(call, task, key)
```
conradbzura: Rename `gen` to `stream`.

wool/src/wool/runtime/worker/process.py
```diff
@@ -36,6 +38,8 @@
 
 logger = logging.getLogger(__name__)
 
+_HAS_UDS = hasattr(socket, "AF_UNIX")
```
conradbzura: Annotate this as `Final[bool]`.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -475,9 +481,9 @@ async def _dispatch(
         return call
 
     async def _execute(
-        self, call: _DispatchCall, task: Task
+        self, call: _DispatchCall, task: Task, key: tuple
```
conradbzura: The `key` parameter is typed as bare `tuple`. Consider using a type alias matching the pool key structure (e.g., `tuple[str, grpc.ChannelCredentials | None, int, WorkerOptions]`) for consistency with `self._key` and to help static analysis catch mismatches.

wool/tests/runtime/worker/test_connection.py
```diff
@@ -1228,6 +1226,69 @@ async def test_dispatch_with_self_dispatch(
...
+            results.append(result)
+
+        # Assert
+        assert results == ["result"]
+        channel_spy.assert_called()
+        uds_calls = [
+            c for c in channel_spy.call_args_list
+            if c.args[0] == uds_target
+        ]
+        assert len(uds_calls) >= 1
```
conradbzura, line 1355: The assertion checks that `insecure_channel` was called with the UDS target but does not verify that the channel was created with `None` credentials (as opposed to the connection's configured credentials). Consider asserting that the UDS-targeted call used `None` credentials to fully cover the behavior described in the PR's test plan. Also consider verifying that a `PassthroughSerializer` was used, since UDS self-dispatch should combine both optimizations.

conradbzura (commented):
`close()` only clears the pool entry for `self._key` (the TCP key). When self-dispatch uses a UDS key, the UDS channel pool entry is never explicitly cleared and must wait for TTL expiry. Consider whether `close()` should also clear the UDS pool entry, or document that the UDS entry is intentionally TTL-managed.

PR #108, opened 2026-03-23
Raise on worker spawn failure in WorkerPool instead of silently discarding exceptions — Closes #107

### LINKED ISSUE #107: Raise on worker spawn failure in WorkerPool instead of silently discarding exceptions
## Description

`WorkerPool._worker_context` uses `asyncio.gather(*tasks, return_exceptions=True)` to start workers concurrently. The `return_exceptions=True` flag causes spawn exceptions to be collected as return values rather than raised. The results are never inspected — the pool proceeds to `yield` regardless of whether any workers actually started.

Since the `gather` is awaited before `yield`, no tasks are in flight when failures occur. The pool simply enters service silently degraded (or with zero workers).

## Expected behavior

If any worker fails to spawn, `WorkerPool` should raise an exception that surfaces the failure(s) to the caller before yielding. Teardown errors (line 388) may remain best-effort since the pool is already shutting down.

## Root cause

`return_exceptions=True` on line 384 of `src/wool/runtime/worker/pool.py`:

```python
await asyncio.gather(*tasks, return_exceptions=True)
yield [w.metadata for w in self._workers if w.metadata]
```

The gathered results are never inspected for exceptions — they are discarded entirely.

conradbzura:
## Summary

Inspect `asyncio.gather` results in `WorkerPool._worker_context` and raise spawn failures instead of silently discarding them. A single failure raises the original exception directly; multiple failures raise an `ExceptionGroup`. The teardown gather remains best-effort since the pool is already shutting down.

Replace the `xfail_known_bugs` infrastructure in the integration test suite with a simpler `retry_grpc_internal` fixture that retries on internal gRPC errors (the PollerCompletionQueue thundering-herd race) and lets all other failures propagate immediately.

Closes #107

### Raise on spawn failure (`pool.py`)

`_worker_context` previously awaited `asyncio.gather(*tasks, return_exceptions=True)` and discarded the results entirely. The pool would enter service silently degraded or with zero workers.

Now the results are inspected after all spawn tasks complete:

```python
results = await asyncio.gather(*tasks, return_exceptions=True)
errors = [r for r in results if isinstance(r, Exception)]
if errors:
    if len(errors) == 1:
        raise errors[0]
    raise ExceptionGroup("worker spawn failures", errors)
```

`return_exceptions=True` is retained so all workers attempt to start before raising — this avoids dangling background tasks. The `try/finally` structure ensures cleanup still runs when the raise occurs before `yield`.

### Replace xfail with retry (integration tests)

Remove the `_KnownBug` dataclass, `_KNOWN_BUGS` registry, and `xfail_known_bugs` fixture. Replace with `retry_grpc_internal` — a fixture that retries test bodies up to 3 times with exponential backoff when the caught exception is an internal gRPC error (`StatusCode.INTERNAL`). All other exceptions propagate immediately. If retries are exhausted, the last exception propagates as a real test failure instead of xfailing.

## Test cases

| 1 | `TestWorkerPool` | A worker that fails during startup | `WorkerPool` is used as async context manager | It should raise `RuntimeError` | Single spawn failure via mock fixture |
| 2 | `TestWorkerPool` | A pool with a failing worker factory | The pool is started | It should raise `RuntimeError` | Single spawn failure via custom factory |
| 3 | `TestWorkerPool` | A pool with size=3 where all workers fail | The pool is used as async context manager | It should raise `ExceptionGroup` with 3 exceptions | Multiple spawn failures |
| 4 | `TestWorkerPool` | A pool with size=3 where 1 of 3 fails | The pool is used as async context manager | It should raise `RuntimeError` for the single failure | Partial spawn failure |

### DIFF
--- a/wool/src/wool/runtime/worker/pool.py
+++ b/wool/src/wool/runtime/worker/pool.py
@@ -381,7 +381,9 @@ class WorkerPool:
             self._workers[worker] = stop(worker)
 
         try:
-            await asyncio.gather(*tasks, return_exceptions=True)
+            results = await asyncio.gather(*tasks, return_exceptions=True)
+            if errors := [r for r in results if isinstance(r, Exception)]:
+                raise ExceptionGroup("worker spawn failures", errors)
             yield [w.metadata for w in self._workers if w.metadata]
         finally:
             tasks = [asyncio.create_task(stop) for stop in self._workers.values()]
--- a/wool/tests/integration/conftest.py
+++ b/wool/tests/integration/conftest.py
@@ -10,7 +10,6 @@ import asyncio
 import datetime
 import ipaddress
 import uuid
-from collections.abc import Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from dataclasses import fields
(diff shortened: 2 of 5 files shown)


wool/src/wool/runtime/worker/pool.py
```diff
@@ -381,7 +381,10 @@ async def stop(worker):
             self._workers[worker] = stop(worker)
 
         try:
-            await asyncio.gather(*tasks, return_exceptions=True)
+            results = await asyncio.gather(*tasks, return_exceptions=True)
+            errors = [r for r in results if isinstance(r, Exception)]
+            if errors:
```
conradbzura: Use the walrus operator to set `errors` here.

PR #117, opened 2026-03-24
Pool discovery protocols shared across proxies on the same worker — Closes #55

### LINKED ISSUE #55: Pool discovery protocols shared across proxies on the same worker
## Summary

When multiple proxies on the same worker use the same discovery protocol (e.g., `LanDiscovery`), each proxy currently starts its own independent discovery instance. Pool discovery protocol instances by its hash (for the current implementations, have hash return the protocol type) so that proxies sharing the same protocol reuse a single underlying subscription, with per-proxy filter predicates overlaid on the shared event stream.

## Motivation

A worker receiving tasks from multiple client proxies that all use `LanDiscovery` will spawn multiple redundant Zeroconf browsers watching the same mDNS traffic. Similarly, multiple `LocalDiscovery` proxies will each attach to the same shared memory region independently. Pooling by protocol type and fanning out filtered views to each proxy's subscriber eliminates this redundancy.

This becomes increasingly important as workers handle tasks from diverse clients — the number of discovery instances scales with the number of distinct proxies rather than staying constant per protocol type.

## Open questions

- Is `__hash__` the correct approach, or should a dedicated reference attribute be added to the structural type for this purpose? Requires some exploration.

conradbzura:
## Summary

Rewrite the subscriber pooling layer so that proxies sharing the same discovery protocol reuse a single underlying subscription with demand-driven multicast. Introduce a `Fanout` utility that wraps any async generator and distributes each item to multiple independent consumers via a leader/follower pull model — no background task required. Late-joining consumers receive a replay of all currently known workers so they start with a consistent view.

All class-level state uses `WeakKeyDictionary` so that entries cascade away automatically when their keys are garbage-collected — no manual cleanup needed beyond closing the async generator on shutdown.

Closes #55

### `Fanout` multicast utility

Add a generic `Fanout[T]` container that wraps a single `AsyncGenerator[T]` source and multicasts each pulled item to every registered `FanoutConsumer`. The first consumer whose queue is empty acquires a shared `asyncio.Lock`, pulls one item from the source via `anext`, distributes it to all other consumers' queues, and returns. Consumers are tracked in a `WeakSet`; when a consumer goes out of scope the weak reference expires and it is silently removed from the fan-out set. `cleanup()` closes the source generator and pushes a sentinel to all remaining consumer queues.

### Demand-driven `_SharedSubscription`

Add a `_SharedSubscription` wrapper that bridges a discovery subscriber and a `Fanout`. Each `__aiter__` call enters a `ResourcePool` resource, lazily wraps the raw subscriber in a shared `Fanout`, and iterates an independent `FanoutConsumer`. Worker state is tracked as a `dict[uid, WorkerMetadata]` per subscriber — only the latest metadata is stored, and entries are removed on `worker-dropped`. When a new consumer registers, it receives a replay of `worker-added` events for all currently known workers.

### `SubscriberMeta` via `__new__`

Switch from `__call__` to `__new__` on the metaclass. The metaclass's `__new__` (called at class definition time) injects a custom `__new__` onto the subscriber class that caches the raw subscriber in the `ResourcePool` and returns a `_SharedSubscription` wrapper. Since the returned object is not an instance of the subscriber class, `type.__call__` skips `__init__` — no double-initialization.

### Pickle support

Add `__reduce__` to `_SharedSubscription` (delegates to the raw subscriber's `__reduce__`, so unpickling goes through the metaclass and re-wraps automatically) and to `afilter` (pickles predicate + inner subscriber).

### Subscriber simplification

Remove the `_consume()` method from both `LanDiscovery.Subscriber` and `LocalDiscovery.Subscriber`. `__aiter__` now returns `_event_stream()` directly. `_shutdown` becomes async and delegates to `_SharedSubscription.cleanup`.

### Integration test: `DURABLE_SHARED` pool mode

Add a new pool arrangement where two `WorkerPool` instances share the same `LocalDiscovery`. The second pool discovers existing workers via late-joiner replay. Include pairwise filter and Hypothesis strategy updates so the new mode is covered across all dimension combinations.

## Test cases

| Test Suite | # | Given | When | Then | Coverage Target |
|---|---|---|---|---|---|
| TestSubscriberMeta | SM-001 | A subscriber class using SubscriberMeta | The class is instantiated | It should return a _SharedSubscription wrapping the cached raw subscriber | __new__ wrapping |
| TestSubscriberMeta | SM-002 | Two construction calls with the same arguments | Instantiated twice | Both should wrap the same underlying raw subscriber | Singleton caching |
| TestSubscriberMeta | SM-003 | Two construction calls with different arguments | Instantiated with different keys | Underlying raw subscribers should be different objects | Key isolation |
| TestSubscriberMeta | SM-004 | The __subscriber_pool__ ContextVar is None | A subscriber is constructed | The ContextVar should be set to a ResourcePool | Lazy pool init |
| TestSharedSubscription | SS-001 | A _SharedSubscription wrapping a source that yields one event | anext is called on the consumer | It should return the event from the source | Demand-driven pull |
| TestSharedSubscription | SS-002 | Two registered consumers sharing the same source | One consumer triggers a pull via anext | Both consumers should receive the event | Fan-out to multiple consumers |
| TestSharedSubscription | SS-003 | A _SharedSubscription wrapping a subscriber | __aiter__ is called twice | It should return two distinct _SharedSubscription instances | Independent consumers |
| TestSharedSubscription | SS-004 | A registered consumer against a subscriber | cleanup is called | Iterator should be removed and sentinel pushed to consumer queues | Cleanup and shutdown |
| TestSharedSubscription | SS-005 | A _SharedSubscription wrapping a stub subscriber with __reduce__ | Pickled and unpickled via cloudpickle | It should produce a _SharedSubscription wrapping a subscriber with the same key | Pickle roundtrip |
| TestSharedSubscription | SS-006 | Two consumers sharing a finite source of 2 events | Both iterate to exhaustion | First consumer should get both events; second should get replay | Source exhaustion + late-joiner replay |
| TestSharedSubscription | SS-007 | A consumer registered against a cleaned-up subscriber | anext is called | It should raise StopAsyncIteration | Post-cleanup termination |
| Testafilter | AF-001 | A source yielding one matching worker-added event | afilter is iterated with an accepting predicate | It should yield the worker-added event | Single matching event |
| Testafilter | AF-002 | A source yielding worker-added then worker-updated | afilter is iterated with an accept-all predicate | It should yield both events in order | Unfiltered passthrough |
| Testafilter | AF-003 | A source yielding a worker-added event | afilter is iterated with a rejecting predicate | It should yield no events | Filter suppression |
| Testafilter | AF-004 | A source yielding worker-added (fails) then worker-updated (passes) | afilter is iterated with a tag filter | It should yield worker-added for the now-matching worker | Filter entry transition |
| Testafilter | AF-005 | A source yielding worker-added (passes) then worker-updated (fails) | afilter is iterated with a tag filter | It should yield worker-added then worker-dropped with prior metadata | Filter exit transition |
| Testafilter | AF-006 | A source yielding worker-added then worker-dropped | afilter is iterated with accept-all | It should yield both events | Tracked worker drop |
| Testafilter | AF-007 | A source yielding worker-added (fails) then worker-dropped | afilter is iterated with rejecting filter | It should yield no events | Untracked worker drop suppression |
| Testafilter | AF-008 | An afilter wrapping a repeatable source | __aiter__ is called twice sequentially | Each iteration should produce independent results | Independent iteration state |
| Testafilter | AF-009 | An afilter wrapping a picklable subscriber | Pickled and unpickled via cloudpickle | It should produce an afilter instance | Pickle roundtrip |
| TestPoolComposition | IC-001 | Two WorkerPool instances sharing the same LocalDiscovery | Both pools entered, coroutine dispatched through primary | It should discover the worker and return the correct result | Shared subscriber end-to-end |

### DIFF
--- a/wool/src/wool/runtime/discovery/__init__.py
+++ b/wool/src/wool/runtime/discovery/__init__.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 from contextvars import ContextVar
+from typing import Any
 from typing import Final
 
 from wool.runtime.resourcepool import ResourcePool
 
-__subscriber_pool__: Final[ContextVar[ResourcePool | None]] = ContextVar(
+__subscriber_pool__: Final[ContextVar[ResourcePool[Any] | None]] = ContextVar(
     "__subscriber_pool__", default=None
 )
--- a/wool/src/wool/runtime/discovery/__init__.py
+++ b/wool/src/wool/runtime/discovery/__init__.py
@@ -1,0 +1,10 @@
+from __future__ import annotations
+
+from contextvars import ContextVar
+from typing import Final
+
+from wool.runtime.resourcepool import ResourcePool
+
+__subscriber_pool__: Final[ContextVar[ResourcePool | None]] = ContextVar(
+    "__subscriber_pool__", default=None
+)
--- a/wool/src/wool/runtime/discovery/lan.py
+++ b/wool/src/wool/runtime/discovery/lan.py
@@ -6,6 +6,7 @@ import json
 import socket
 from asyncio import Queue
 from types import MappingProxyType
+from typing import AsyncGenerator
 from typing import AsyncIterator
 from typing import Dict
 from typing import Final
@@ -27,6 +28,8 @@ from wool.runtime.discovery.base import DiscoveryPublisherLike
 from wool.runtime.discovery.base import DiscoverySubscriberLike
 from wool.runtime.discovery.base import PredicateFunction
 from wool.runtime.discovery.base import WorkerMetadata
+from wool.runtime.discovery.pool import SubscriberMeta
+from wool.utilities.afilter import afilter
 
 
 # public
@@ -85,6 +88,14 @@ class LanDiscovery(Discovery):
         self._filter = filter
         self._service_type = _namespaced_service_type(self._namespace)
 
+    def __hash__(self) -> int:
+        return hash((type(self), self._namespace))
+
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, LanDiscovery):
+            return self._namespace == other._namespace
+        return NotImplemented
+
     @property
     def namespace(self) -> str:
         """The namespace identifier for this discovery service.
@@ -127,10 +138,11 @@ class LanDiscovery(Discovery):
             A subscriber instance that receives filtered worker
             discovery events.
         """
-        return self.Subscriber(
-            self._service_type,
-            filter if filter is not None else self._filter,
-        )
+        effective = filter if filter is not None else self._filter
+        subscriber = self.Subscriber(self._service_type)
+        if effective is not None:
+            return afilter(effective, subscriber)
+        return subscriber
 
     class Publisher:
         """Publisher for broadcasting worker discovery events.
@@ -311,7 +323,7 @@ class LanDiscovery(Discovery):
 
             return socket.inet_aton(socket.gethostbyname(host)), port
 
-    class Subscriber:
+    class Subscriber(metaclass=SubscriberMeta):
         """Subscriber for receiving worker discovery events.
 
         Subscribes to worker :class:`discovery events
@@ -319,36 +331,33 @@ class LanDiscovery(Discovery):
         local network. As workers register and unregister their
         services, the subscriber yields corresponding events.
 
-        Each call to ``__aiter__`` creates an isolated iterator with its
-        own state. Multiple concurrent iterations from the same
-        subscriber instance are fully independent.
+        Each call to ``__aiter__`` creates an isolated consumer that
+        receives events from a shared underlying Zeroconf browser via
+        :class:`~wool.runtime.discovery.pool._SharedSubscription`.
+        Multiple concurrent iterations are fully independent.
 
-        Uses AsyncZeroconf's service browser to monitor for service
-        changes and converts Zeroconf events into Wool discovery
-        events.
+        Instances are cached as singletons by
+        :class:`~wool.runtime.discovery.pool.SubscriberMeta` — two
+        calls with the same ``service_type`` return the same object.
 
         :param service_type:
             The DNS-SD service type string for this namespace.
-        :param filter:
-            Optional predicate function to filter workers. Only workers
-            for which the predicate returns True will be included in
-            events.
         """
 
-        _filter: Final[PredicateFunction[WorkerMetadata] | None]
-
-        def __init__(
-            self,
-            service_type: str,
-            filter: PredicateFunction[WorkerMetadata] | None = None,
-        ) -> None:
+        def __init__(self, service_type: str) -> None:
             self.service_type = service_type
-            self._filter = filter
+
+        @classmethod
+        def _cache_key(cls, service_type: str) -> tuple:
+            return (cls, service_type)
+
+        async def _shutdown(self) -> None:
+            """Clean up shared subscription state for this subscriber."""
 
         def __aiter__(self) -> AsyncIterator[DiscoveryEvent]:
             return self._event_stream()
 
-        async def _event_stream(self) -> AsyncIterator[DiscoveryEvent]:
+        async def _event_stream(self) -> AsyncGenerator[DiscoveryEvent, None]:
             """Stream discovery events from the network.
 
             Creates isolated state for this iteration including its own
@@ -357,13 +366,13 @@ class LanDiscovery(Discovery):
             completes or is interrupted.
 
             :yields:
-                Discovery events as workers are added, updated, or removed.
+                Discovery events as workers are added, updated, or
+                removed.
             """
-            # Create isolated state for this iterator
             event_queue: Queue[DiscoveryEvent] = Queue()
             service_cache: Dict[str, WorkerMetadata] = {}
 
-            # Configure zeroconf to use localhost only to avoid network warnings
+            # Configure zeroconf to use localhost only
             aiozc = AsyncZeroconf(interfaces=["127.0.0.1"])
 
             try:
@@ -375,7 +384,6 @@ class LanDiscovery(Discovery):
                         aiozc=aiozc,
                         event_queue=event_queue,
                         service_cache=service_cache,
-                        predicate=self._filter or (lambda _: True),
                     ),
                 )
 
@@ -401,13 +409,10 @@ class LanDiscovery(Discovery):
             :param service_cache:
                 Cache to track service properties for pre/post event
                 states.
-            :param predicate:
-                Function to filter which workers to track.
             """
 
             aiozc: AsyncZeroconf
             _event_queue: Queue[DiscoveryEvent]
-            _service_addresses: Dict[str, str]
             _service_cache: Dict[str, WorkerMetadata]
 
             def __init__(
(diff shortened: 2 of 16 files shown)


wool/src/wool/runtime/discovery/lan.py
```diff
@@ -346,19 +354,38 @@ def __init__(
             self._filter = filter
 
         def __aiter__(self) -> AsyncIterator[DiscoveryEvent]:
-            return self._event_stream()
+            from wool.runtime.discovery.pool import subscribe_pooled
```
conradbzura: Move this to the top of the module.

wool/src/wool/runtime/discovery/lan.py
```diff
@@ -346,19 +354,38 @@ def __init__(
...
+            # Key by service_type (derived deterministically from namespace)
+            # because the Subscriber does not store the namespace directly.
+            key = (type(self), self.service_type)
+            return subscribe_pooled(
+                key,
+                source_factory=lambda: self._event_stream(
+                    predicate=lambda _: True,
+                ),
+                filter=self._filter,
+            )
```
conradbzura: This is getting messy - we should keep this untouched and instead add a general-purpose `afilter` utility function that accepts the predicate function and returns a filtered iterator. This decouples filtering from discovery.

wool/src/wool/runtime/discovery/lan.py
conradbzura: Once https://github.com/wool-labs/wool/pull/117/changes#r2983324448 is addressed, we will have an `afilter` utility function we can use here instead of passing the filter to `Subscriber`.

wool/src/wool/runtime/discovery/pool.py
conradbzura, line 1: Once https://github.com/wool-labs/wool/pull/117/changes#r2983324448 is addressed, this will become largely unnecessary. We can just use a naive `ResourcePool` since we won't need to modify the stream with filters at this level - the discovery protocol itself will wrap it, or users can wrap the subscriber with `afilter` directly.

wool/src/wool/runtime/discovery/lan.py
conradbzura: Once https://github.com/wool-labs/wool/pull/117/changes#r2983324448 is addressed, we can make this a singleton with `ResourcePool` as the cache. Add a `SubscriberMeta` metaclass that intercepts construction and returns the cached subscriber from the global resource pool.

wool/src/wool/runtime/discovery/local.py
conradbzura: See https://github.com/wool-labs/wool/pull/117/changes#r2983367114 - the singleton metaclass pattern should be implemented here as well.

wool/src/wool/runtime/discovery/lan.py
conradbzura, line 1: Once https://github.com/wool-labs/wool/pull/117/changes#r2983367114 is addressed, the `discovery` subpackage should initialize a global `__subscriber_pool__: ContextVar[ResourcePool | None]` variable to None. The proposed `SubscriberMeta` class should lazily set the var to an empty subscriber `ResourcePool` if it's None, and use it to get a cached subscriber.

wool/src/wool/runtime/worker/service.py
```diff
@@ -432,6 +432,9 @@ async def _stop(self, *, timeout: float | None = 0) -> None:
         try:
             if proxy_pool := wool.__proxy_pool__.get():
                 await proxy_pool.clear()
+            from wool.runtime.discovery.pool import reset as _reset_discovery_pool
+
+            await _reset_discovery_pool()
```
conradbzura: Once https://github.com/wool-labs/wool/pull/117/changes#r2983408882 is addressed, this will need to clear the subscriber pool if one exists.

wool/src/wool/runtime/discovery/lan.py
```diff
@@ -311,44 +324,59 @@ def _resolve_address(self, address: str) -> Tuple[bytes, int]:
...
             self.service_type = service_type
-            self._filter = filter
+
+        @classmethod
+        def _cache_key(cls, service_type: str) -> tuple:
+            return (cls, service_type)
+
+        def _shutdown(self) -> None:
+            """No-op — shared subscriptions are cleaned up per
+            iteration in :meth:`_consume`."""
```
conradbzura: What's this for if it doesn't do anything...

wool/src/wool/runtime/discovery/__init__.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import annotations
+
+from contextvars import ContextVar
+from typing import Final
+
+from wool.runtime.resourcepool import ResourcePool
+
+__subscriber_pool__: Final[ContextVar[ResourcePool | None]] = ContextVar(
```
conradbzura: Add a type to `ResourcePool`.

wool/src/wool/runtime/discovery/lan.py
```diff
@@ -311,44 +324,59 @@ def _resolve_address(self, address: str) -> Tuple[bytes, int]:
...
         def __aiter__(self) -> AsyncIterator[DiscoveryEvent]:
-            return self._event_stream()
+            return self._consume()
+
+        async def _consume(self) -> AsyncIterator[DiscoveryEvent]:
+            """Add a consumer to the shared subscription and yield
+            events from its queue."""
+            shared = _SharedSubscription(
+                source_factory=lambda: self._event_stream(),
+            )
```
conradbzura: Why is this getting wrapped here? The metaclass should implement `__new__` and wrap the cached subscriber there. _SharedSubscription should track state, i.e. map a new consumer to the wrapped subscriber. The mapping should be a weak value dictionary so that consumers that go out of scope are cleaned up automatically.

wool/src/wool/runtime/discovery/lan.py
```diff
@@ -311,44 +324,42 @@ def _resolve_address(self, address: str) -> Tuple[bytes, int]:
...
             self.service_type = service_type
-            self._filter = filter
+
+        @classmethod
+        def _cache_key(cls, service_type: str) -> tuple:
+            return (cls, service_type)
+
+        async def _shutdown(self) -> None:
+            """Clean up shared subscription state for this subscriber."""
+            await _SharedSubscription.cleanup(self)
```
conradbzura: Is this used anywhere?

wool/src/wool/runtime/discovery/pool.py
```diff
@@ -0,0 +1,227 @@
...
+        mcs,
+        name: str,
+        bases: tuple[type, ...],
+        namespace: dict[str, Any],
+    ) -> SubscriberMeta:
+        cls = super().__new__(mcs, name, bases, namespace)
+        original_init = cls.__init__  # type: ignore[misc]
+
+        def _subscriber_new(cls_arg: type, *args: Any, **kwargs: Any) -> Any:
+            key = cls_arg._cache_key(*args, **kwargs)  # type: ignore[attr-defined]
```
conradbzura, line 126: This should be a kwarg to `__new__` so that usage can be `FooSub(metaclass=SubscriberMeta, key=...)`.

wool/src/wool/runtime/discovery/pool.py
```diff
@@ -0,0 +1,227 @@
...
+                    DiscoveryEvent("worker-added", metadata=metadata)
+                )
+
+    def __reduce__(self) -> tuple:
+        # Delegate to the raw subscriber — unpickling goes through
+        # SubscriberMeta which re-wraps in _SharedSubscription.
+        return self._subscriber.__reduce__()
+
+    def __aiter__(self) -> _SharedSubscription:
+        return _SharedSubscription(self._subscriber)
```
conradbzura: Why not return `self`?

Also, we need to wrap the return value with an async generator that enters the `wool/src/wool/resourcepool.py::Reference` context before starting.

PR #119, opened 2026-03-25
Remove parent-resolution from wool.routine wrapper — Closes #118

### LINKED ISSUE #118: Remove parent-resolution from wool.routine wrapper
## Description

`_resolve()` in `wool.runtime.routine.wrapper` walks `sys.modules[fn.__module__]` to re-resolve the decorated function from the module namespace on every invocation. For nested routine dispatch (routine A on Worker A dispatches routine B to Worker B), this causes the dispatched callable to come from the worker's re-imported `__mp_main__` module rather than from the cloudpickle-deserialized closure. Exception classes defined alongside the routines end up as distinct class objects on the caller vs the callee, so `except CustomError` silently fails to catch the re-raised exception.

The bug is observable in any nested `@wool.routine` scenario where a worker-side `except` clause targets a user-defined exception that was raised by a routine dispatched to a different worker.

## Root cause

`_resolve()` resolves `fn` by walking `sys.modules[fn.__module__]` using `fn.__qualname__`. On a spawned worker the re-imported module lives under `__mp_main__`, producing a **different** function object whose globals reference **different** class objects. When `_dispatch` serializes this re-resolved function with cloudpickle, the target worker deserializes it with the re-imported class hierarchy. The calling worker's `except` clause still holds the cloudpickle-deserialized class from the original `__main__` — a different identity — so `isinstance` fails and the exception propagates uncaught.

`_resolve` also introduces unnecessary coupling: it requires the function to be accessible from its `__module__` at call time, which fails for scripts run via `python -c` and other environments where `__main__` is not re-importable.

## Expected behavior

- Nested routine dispatch preserves exception class identity: `except UserDefinedException` on a calling worker catches exceptions raised by a dispatched routine on a different worker.
- `_resolve` and its associated classmethod/staticmethod descriptor handling in `_stream` and `_execute` are removed. `@wool.routine` receives the raw function directly; `@classmethod`/`@staticmethod` must be applied as the outer decorator.

conradbzura:
## Summary

Remove `_resolve()` and its associated `classmethod`/`staticmethod` descriptor handling from `wool.routine`. The decorator now operates directly on the raw function passed to it, eliminating the per-invocation `sys.modules` walk that produced distinct class objects across workers and broke exception identity in nested dispatch scenarios.

`@wool.routine` now raises `ValueError` when applied to a `classmethod` or `staticmethod` descriptor. The correct decorator order is `@classmethod @wool.routine` (descriptor outer, routine inner). An info admonition documenting this requirement is added to the `routine()` docstring.

Closes #118

### Remove _resolve and simplify dispatch internals

`_resolve()` walked `sys.modules[fn.__module__]` using `fn.__qualname__` to re-resolve the decorated function on every call. On spawned workers the re-imported module lives under `__mp_main__`, producing a different function object whose globals reference different class objects — breaking `isinstance` checks and `except` clauses for user-defined exceptions in nested dispatch.

With `_resolve` removed, the wrappers pass `fn` (the original closure-captured function) directly to `_dispatch`, `_stream`, and `_execute`. The `parent` parameter and the `classmethod`/`staticmethod` branch logic in `_stream` and `_execute` are also removed since the decorator now always receives the raw function.

### Add descriptor guard with clear error message

A new guard at the top of `routine()` raises `ValueError("@wool.routine must be applied before @classmethod/@staticmethod")` when passed a descriptor object. This gives an actionable error message instead of the generic "Expected a coroutine function" that would otherwise fire.

### Update decorator order in test fixtures and integration routines

All `@routine @classmethod` / `@routine @staticmethod` usages in test fixtures and integration routine definitions are flipped to the correct `@classmethod @routine` / `@staticmethod @routine` order.

## Test cases

| 1 | (module) | `@routine` applied to a classmethod descriptor | Decorator is evaluated | `ValueError` with message about correct order | Descriptor guard (classmethod) |
| 2 | (module) | `@routine` applied to a staticmethod descriptor | Decorator is evaluated | `ValueError` with message about correct order | Descriptor guard (staticmethod) |
| 3 | (module) | `@routine` on various function types with dispatch enabled | Decorated function is called | Task is dispatched and result returned correctly | Dispatch path for all 8 function types |
| 4 | (module) | `@routine` on various function types with dispatch disabled | Decorated function is called | Function executes locally and result returned correctly | Local execution path for all 8 function types |

### DIFF
--- a/wool/src/wool/runtime/routine/wrapper.py
+++ b/wool/src/wool/runtime/routine/wrapper.py
@@ -6,14 +6,9 @@ from inspect import getsourcelines
 from inspect import isasyncgen
 from inspect import isasyncgenfunction
 from inspect import iscoroutinefunction
-from sys import modules
-from types import ModuleType
 from typing import TYPE_CHECKING
 from typing import AsyncGenerator
 from typing import Coroutine
-from typing import ParamSpec
-from typing import Tuple
-from typing import Type
 from typing import TypeVar
 from typing import cast
 from uuid import uuid4
@@ -27,8 +22,6 @@ if TYPE_CHECKING:
     from wool.runtime.worker.proxy import WorkerProxy
 
 C = TypeVar("C", bound=Callable[..., Coroutine | AsyncGenerator])
-P = ParamSpec("P")
-R = TypeVar("R")
 
 
 # public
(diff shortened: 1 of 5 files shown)


wool/src/wool/runtime/routine/wrapper.py
```diff
@@ -47,15 +40,35 @@ def routine(fn: C) -> C:
         The decorated function that dispatches to the worker pool when
         called.
     :raises ValueError:
-        If the decorated function is not a coroutine function or async
-        generator function.
+        If ``fn`` is a ``classmethod`` or ``staticmethod`` descriptor
```
conradbzura: "``fn``" should be "decorated function".

PR #133, opened 2026-03-27
Add upper-limit support to WorkerPool size — Closes #132

### LINKED ISSUE #132: Add upper-limit to WorkerPool size
## Description

Accept `int | tuple[int, int]` for the `size` parameter on `WorkerPool`. When a tuple is provided, the first element is the number of workers to spawn (as today) and the second element is an upper limit on total pool size — capping how many discovered workers the pool will accept. When `size` is a plain `int`, behavior is unchanged (no cap on discovered workers).

## Motivation

In hybrid pools (spawned workers + discovery), there is currently no way to limit the total number of workers the pool will use. Discovery can add an unbounded number of remote workers, which may not be desirable when resource constraints or cost controls apply. An upper limit gives operators a simple knob to bound pool growth without filtering at the discovery layer.

## Expected outcome

- `WorkerPool(size=4)` behaves exactly as it does today — spawns 4 workers, no cap on discovered workers.
- `WorkerPool(size=(4, 8), discovery=...)` spawns 4 local workers and accepts up to 4 additional discovered workers (8 total).
- The `size` parameter type across all three `__init__` overloads reflects `int | tuple[int, int]`.
- Validation rejects tuples where the upper limit is less than the spawn count.

conradbzura:
## Summary

Add a `lease` parameter to `WorkerPool` that caps how many additional discovered workers the pool will accept. The total pool capacity is `size + lease` when both are set, or just `lease` for durable (discovery-only) pools. Zero is allowed when `size` is set (caps at exactly the spawn count) but rejected for discovery-only pools. The lease count is a cap on admission, not a reservation — discovered workers may serve multiple pools simultaneously, and there is no guarantee that a leased slot will remain filled for the life of the pool. Omitting `lease` preserves existing behavior (unbounded).

The cap is enforced at the `WorkerProxy` sentinel level — the single point through which all discovered workers flow into the loadbalancer. `WorkerProxy` enforces `lease > 0` independently since it always receives the computed total (`size + lease`), never the raw zero. The sentinel guards `worker-updated` events for never-added workers to prevent stale updates leaking through the cap. The `lease` value survives pickling via `__reduce__` so the cap is preserved across nested routine dispatch.

Size resolution is extracted to a typed module-level `_resolve_size` helper to eliminate duplicated validation across match arms.

Closes #132

### Lease parameter and validation (pool.py)

Add `lease: int | None = None` to all three `__init__` overloads and the implementation signature. Reject negative lease values universally; reject zero lease only for discovery-only pools where it would be nonsensical. Extract a `_resolve_size(size: int) -> int` module-level helper that resolves `size=0` to CPU count and validates `size >= 0`. Compute `max_workers = size + lease` in each match arm after size resolution, forwarding the total to `WorkerProxy` in all four branches — including the durable branch which passes the raw `lease` since there is no spawn count to add.

### Cap enforcement in the worker sentinel (proxy.py)

Add a `lease: int | None` parameter to `WorkerProxy` with its own `< 1` validation and `:param` documentation. In the sentinel, skip `"worker-added"` discovery events when the loadbalancer already holds `>= lease` workers. Guard `"worker-updated"` events so updates for never-added workers (rejected by the cap) are silently dropped rather than passed to the loadbalancer. Include `lease` in `__reduce__` so the cap survives pickling for nested routine dispatch.

### Documentation (README.md, src/wool/runtime/worker/README.md)

Update both the root README and worker subpackage README to explain the `lease` parameter with code examples for hybrid and durable pool modes. Clarify that the lease count is a cap on admission, not a reservation.

## Test cases

| 1 | `TestWorkerPool` | A size of 4 and lease of 8 | WorkerPool is instantiated | Creates pool successfully | Lease accepted with valid size |
| 2 | `TestWorkerPool` | A discovery service and lease of 8 with no size | WorkerPool is instantiated | Creates pool successfully | Durable mode with lease |
| 3 | `TestWorkerPool` | A size of 4 and lease of 0 | WorkerPool is instantiated | Creates pool successfully | Zero lease with size accepted |
| 4 | `TestWorkerPool` | A discovery service and lease of 0 with no size | WorkerPool is instantiated | Raises ValueError | Zero lease discovery-only rejected |
| 5 | `TestWorkerPool` | A negative lease | WorkerPool is instantiated | Raises ValueError | Negative lease rejected |
| 6 | `TestWorkerPool` | Size=2, lease=4, and discovery (hybrid mode) | Pool context is entered | WorkerProxy receives `lease=6` (size + lease) | Hybrid mode forwarding |
| 7 | `TestWorkerPool` | Size=2 and lease=4 without discovery (ephemeral) | Pool context is entered | WorkerProxy receives `lease=6` (size + lease) | Ephemeral mode forwarding |
| 8 | `TestWorkerPool` | Discovery only and lease=6 (durable mode) | Pool context is entered | WorkerProxy receives `lease=6` | Durable mode forwarding |
| 9 | `TestWorkerPool` | No explicit size or discovery, lease=4 (default) | Pool context is entered | WorkerProxy receives `lease=cpu_count+4` | Default mode forwarding |
| 10 | `TestWorkerPool` | Plain int size with no lease | Pool context is entered | WorkerProxy receives `lease=None` | No lease forwards None |
| 11 | `TestWorkerPool` | Any positive size and any non-negative lease | WorkerPool is instantiated | Creates pool successfully | Valid combinations accepted |
| 12 | `TestWorkerPool` | Any negative lease value | WorkerPool is instantiated | Raises ValueError | Negative lease rejected |
| 13 | `TestWorkerProxy` | Discovery and `lease=0` | WorkerProxy is instantiated | Raises ValueError | Zero lease rejected by proxy |
| 14 | `TestWorkerProxy` | Discovery and `lease=-1` | WorkerProxy is instantiated | Raises ValueError | Negative lease rejected by proxy |
| 15 | `TestWorkerProxy` | Discovery and `lease=5` | WorkerProxy is instantiated | Creates proxy successfully | Positive lease accepted by proxy |
| 16 | `TestWorkerProxy` | `lease=2` and 3 worker-added events | Proxy started, discovery drains | Only 2 workers accepted | Sentinel cap enforcement |
| 17 | `TestWorkerProxy` | `lease=None` and 3 worker-added events | Proxy started, discovery drains | All 3 workers accepted | No cap when None |
| 18 | `TestWorkerProxy` | `lease=2` at capacity, worker-updated event | Proxy started, discovery drains | Update processed without blocking | Updates bypass cap |
| 19 | `TestWorkerProxy` | `lease=2` at capacity, drop then add | Proxy started, discovery drains | New worker accepted after drop | Drop frees lease slot |
| 20 | `TestWorkerProxy` | `lease=1`, rejected worker then updated | Proxy started, discovery drains | Update for rejected worker silently dropped | Guard on unknown updates |
| 21 | `TestWorkerProxy` | WorkerProxy with `lease=2` | Pickle round-trip then start restored proxy | Only 2 workers accepted by restored proxy | Cap survives pickling |

### DIFF
--- a/wool/src/wool/runtime/worker/pool.py
+++ b/wool/src/wool/runtime/worker/pool.py
@@ -137,6 +137,10 @@ class WorkerPool:
         Capability tags for spawned workers.
     :param size:
         Number of workers to spawn (0 = CPU count).
+    :param lease:
+        Maximum number of additionally discovered workers to admit to the pool.
+        The total pool capacity is ``size + lease`` when both are set, or just
+        ``lease`` for external pools. Defaults to ``None`` (unbounded).
     :param worker:
         Worker factory callable. Defaults to :class:`LocalWorker`.
     :param loadbalancer:
@@ -163,6 +167,7 @@ class WorkerPool:
         self,
         *tags: str,
         size: int = 0,
+        lease: int | None = None,
         worker: WorkerFactory = LocalWorker,
         discovery: None = None,
         loadbalancer: (
@@ -180,6 +185,7 @@ class WorkerPool:
     def __init__(
         self,
         *,
+        lease: int | None = None,
         discovery: DiscoveryLike | Factory[DiscoveryLike],
         loadbalancer: (
             LoadBalancerLike | Factory[LoadBalancerLike]
@@ -197,6 +203,7 @@ class WorkerPool:
         self,
         *tags: str,
         size: int = 0,
+        lease: int | None = None,
         worker: WorkerFactory = LocalWorker,
         discovery: DiscoveryLike | Factory[DiscoveryLike],
         loadbalancer: (
@@ -214,6 +221,7 @@ class WorkerPool:
         self,
         *tags: str,
         size: int | None = None,
+        lease: int | None = None,
         worker: WorkerFactory | None = None,
         discovery: DiscoveryLike | Factory[DiscoveryLike] | None = None,
         loadbalancer: (
@@ -224,15 +232,13 @@ class WorkerPool:
         self._workers = {}
         self._credentials = credentials
 
+        if lease is not None and lease < 0:
+            raise ValueError("Lease must be non-negative")
+
         match (size, discovery):
             case (size, discovery) if size is not None and discovery is not None:
-                if size == 0:
-                    cpu_count = os.cpu_count()
-                    if cpu_count is None:
-                        raise ValueError("Unable to determine CPU count")
-                    size = cpu_count
-                elif size < 0:
-                    raise ValueError("Size must be non-negative")
+                size = _resolve_size(size)
+                max_workers = size + lease if lease is not None else None
 
                 @asynccontextmanager
                 async def create_proxy():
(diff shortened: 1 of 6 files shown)


wool/src/wool/runtime/worker/pool.py
```diff
@@ -253,19 +270,14 @@ async def create_proxy():
                                 discovery=discovery_svc.subscribe(_predicate(tags)),
                                 loadbalancer=loadbalancer,
                                 credentials=self._credentials,
+                                max_workers=limit,
```
conradbzura: Let's call this `limit` to mirror `WorkerPool`.

PR #137, opened 2026-03-28
Add PEP 702 static-analysis deprecation overloads for `size` parameter — Closes #136

### LINKED ISSUE #136: Add PEP 702 static-analysis deprecation overloads for `size` parameter
## Description

Add two additional `@overload` declarations to `WorkerPool.__init__` decorated with `@deprecated("Use 'spawn' instead of 'size'.")` from `typing_extensions` — one for ephemeral mode and one for hybrid mode. Type checkers will flag `WorkerPool(size=...)` at static analysis time, complementing the existing runtime `DeprecationWarning`.

## Motivation

The `size` parameter was deprecated at runtime in #135, but static analysis tools (pyright, mypy) have no way to surface the deprecation. PEP 702 introduced `warnings.deprecated` (backported in `typing_extensions`) specifically for this purpose. Adding overloads with the `@deprecated` decorator closes the gap so developers get IDE hints and type-checker warnings before they ever run the code.

## Expected outcome

- Two new `@overload` signatures on `WorkerPool.__init__` covering `size=` usage (ephemeral and hybrid modes) are decorated with `@deprecated("Use 'spawn' instead of 'size'.")`.
- Type checkers (pyright, mypy) emit a deprecation diagnostic when `WorkerPool(size=...)` is used.
- The existing runtime `DeprecationWarning` remains unchanged.

conradbzura:
## Summary

Add two `@overload` declarations decorated with `@deprecated` from `typing_extensions` to `WorkerPool.__init__`, covering ephemeral and hybrid modes. Type checkers (pyright, mypy) now emit a deprecation diagnostic when `WorkerPool(size=...)` appears in source, complementing the runtime `DeprecationWarning` introduced in #135.

Closes #136

### Deprecated overloads for `size` parameter

Import `deprecated` from `typing_extensions` and add two new `@overload` stubs after the existing three overloads on `WorkerPool.__init__`:

- **Ephemeral mode** — `size: int` with `discovery: None` (mirrors the existing ephemeral `spawn` overload).
- **Hybrid mode** — `size: int` with `discovery: DiscoveryLike | Factory[DiscoveryLike]` (mirrors the existing hybrid `spawn` overload).

Both are decorated with `@deprecated("Use 'spawn' instead of 'size'.")`. The durable mode has no `spawn`/`size` parameter, so no deprecated overload is needed there.

`size` is declared as a required `int` (no default) because callers always pass it explicitly — a default would weaken the overload's discriminating power and match calls that should resolve to the non-deprecated `spawn` overloads.

No runtime behavior changes. The `@deprecated` decorator on `@overload` stubs is a static-analysis-only annotation; the existing runtime deprecation path is untouched.

### DIFF
--- a/wool/src/wool/runtime/worker/pool.py
+++ b/wool/src/wool/runtime/worker/pool.py
@@ -13,6 +13,8 @@ from typing import Coroutine
 from typing import Final
 from typing import overload
 
+from typing_extensions import deprecated
+
 from wool.runtime.discovery.base import DiscoveryLike
 from wool.runtime.discovery.base import DiscoveryPublisherLike
 from wool.runtime.discovery.local import LocalDiscovery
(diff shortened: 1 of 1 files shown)


conradbzura (commented):
No issues found. The diff is clean — overload stacking order, import placement, and signature design all look correct.

PR #140, opened 2026-03-28
Accept cache key callable as a SubscriberMeta keyword argument — Closes #120

### LINKED ISSUE #120: Accept cache key callable as a SubscriberMeta keyword argument
## Description

Replace the `_cache_key` classmethod convention on subscriber classes with a `key` keyword argument passed at class definition time via `metaclass=SubscriberMeta, key=...`. The metaclass's `__new__` should accept `key` as a `**kwargs` entry, store it on the class, and use it in the injected `__new__` to compute the `ResourcePool` cache key. Subscriber classes should no longer need to define `_cache_key` themselves.

Current usage:

```python
class Subscriber(metaclass=SubscriberMeta):
    @classmethod
    def _cache_key(cls, namespace):
        return (cls, namespace)
```

Target usage:

```python
class Subscriber(metaclass=SubscriberMeta, key=lambda cls, namespace: (cls, namespace)):
    ...
```

## Motivation

The `_cache_key` classmethod is a protocol convention that every subscriber must implement but nothing enforces at definition time. Passing it as a keyword argument to the metaclass makes the requirement explicit, moves the configuration closer to the declaration site, and eliminates the need for a magic-named classmethod.

## Expected outcome

`SubscriberMeta.__new__` accepts a `key` keyword argument. Subscriber classes pass their cache key callable via `metaclass=SubscriberMeta, key=...`. The `_cache_key` classmethod convention is removed from all subscriber implementations.

conradbzura:
## Summary

Replace the `_cache_key` classmethod convention on subscriber classes with a `key` keyword argument passed at class definition time. `SubscriberMeta.__new__` now pops `key` from the metaclass keyword arguments, stores it on the class, and invokes it in the injected `__new__` to compute the `ResourcePool` cache key. The `_cache_key` classmethod is removed from all subscriber implementations.

Closes #120

### Accept `key` kwarg in `SubscriberMeta.__new__`

Add `**kwargs` to the metaclass `__new__` signature and pop a `key` entry before forwarding remaining kwargs to `type.__new__`. When present, store the callable on the class as `_cache_key_fn`. The injected `_subscriber_new` calls `cls_arg._cache_key_fn(cls_arg, *args, **kwargs)` instead of the old `cls_arg._cache_key(...)`.

The `key` callable is a plain function (not a classmethod), so `cls` must be passed explicitly as the first argument. This matches the convention of `sorted(key=...)` and similar stdlib APIs.

Before:
```python
class Subscriber(metaclass=SubscriberMeta):
    @classmethod
    def _cache_key(cls, namespace):
        return (cls, namespace)
```

After:
```python
class Subscriber(metaclass=SubscriberMeta, key=lambda cls, namespace: (cls, namespace)):
    ...
```

### Update subscriber classes

Remove the `_cache_key` classmethod from `LocalDiscovery.Subscriber` and `LanDiscovery.Subscriber`, passing the equivalent lambda via `key=` in their class definitions.

### Update and add tests

Update existing test stubs to use `key=` instead of `_cache_key`. Add two new tests: one verifying the key callable receives `cls` as its first argument, another verifying subclass inheritance of the parent's key callable.

## Test cases

|---|---|---|---|---|---|
| 1 | `TestSubscriberMeta` | A subscriber class using SubscriberMeta with `key=` | The class is instantiated | Returns a `_SharedSubscription` | Metaclass wrapping |
| 2 | `TestSubscriberMeta` | A source yielding events and two subscriptions with the same key | Both consumers iterate | Events are fanned out to both | Shared fan-out |
| 3 | `TestSubscriberMeta` | Two subscriptions with different keys | Both are iterated | Events are isolated per key | Key isolation |
| 4 | `TestSubscriberMeta` | The subscriber pool ContextVar is None | A subscriber is constructed | The ContextVar is set to a ResourcePool | Lazy pool init |
| 5 | `TestSubscriberMeta` | A subscriber class with a key callable that captures cls | The class is instantiated | The key callable receives the class as its first argument | Explicit cls passing |
| 6 | `TestSubscriberMeta` | A parent class with `key=` and a subclass without | The subclass is instantiated | It inherits the parent's key callable | Key inheritance |

### DIFF
--- a/wool/src/wool/runtime/discovery/lan.py
+++ b/wool/src/wool/runtime/discovery/lan.py
@@ -323,7 +323,10 @@ class LanDiscovery(Discovery):
 
             return socket.inet_aton(socket.gethostbyname(host)), port
 
-    class Subscriber(metaclass=SubscriberMeta):
+    class Subscriber(
+        metaclass=SubscriberMeta,
+        key=lambda cls, service_type: (cls, service_type),
+    ):
         """Subscriber for receiving worker discovery events.
 
         Subscribes to worker :class:`discovery events
@@ -346,10 +349,6 @@ class LanDiscovery(Discovery):
 
         def __init__(self, service_type: str) -> None:
             self.service_type = service_type
-
-        @classmethod
-        def _cache_key(cls, service_type: str) -> tuple:
-            return (cls, service_type)
 
         async def _shutdown(self) -> None:
             """Clean up shared subscription state for this subscriber."""
--- a/wool/src/wool/runtime/discovery/local.py
+++ b/wool/src/wool/runtime/discovery/local.py
@@ -560,7 +560,14 @@ class LocalDiscovery(Discovery):
                 pass  # pragma: no cover
             atexit.unregister(self._cleanups.pop(shared_memory.name))
 
-    class Subscriber(metaclass=SubscriberMeta):
+    class Subscriber(
+        metaclass=SubscriberMeta,
+        key=lambda cls, namespace, *, poll_interval=None: (
+            cls,
+            namespace,
+            poll_interval,
+        ),
+    ):
         """Subscriber for receiving worker discovery events.
 
         Subscribes to worker :class:`discovery events <~wool.DiscoveryEvent>`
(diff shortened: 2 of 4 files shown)


conradbzura (commented):
No issues found. All guide checks pass — naming, docstrings, coverage, and correctness are clean.

PR #143, opened 2026-03-29
Make worker proxy start lazily on dispatch by default — Closes #54

### LINKED ISSUE #54: Make worker proxy start lazily on dispatch by default
## Summary

Defer proxy instantiation on worker processes until a nested `@wool.routine` is actually invoked. Currently, the deserialized proxy is eagerly started when a task arrives on a worker — this triggers a discovery subscription and sentinel task even if the task never dispatches nested routines. Change this so the proxy is only instantiated if and when the task's callable dispatches a nested routine.

## Motivation

Most tasks don't call nested routines — they execute their callable and return a result. Eagerly starting a proxy for every inbound task wastes resources: it subscribes to discovery (e.g., spawning a Zeroconf browser or attaching to shared memory), starts a sentinel task to track worker events, and allocates a load balancer context. gRPC connections are already lazy, but the discovery overhead still compounds on workers handling high task throughput with no nesting.

Lazy instantiation ensures workers only pay the cost of discovery and sentinel setup when they actually need to dispatch sub-tasks.

## Affected code

- `wool/src/wool/runtime/worker/service.py` — proxy setup in task execution path
- `wool/src/wool/runtime/worker/proxy.py` — proxy lifecycle management

conradbzura:
## Summary

Add a `lazy` parameter (default `True`) to `WorkerProxy` and `WorkerPool` that defers discovery subscription, sentinel task creation, and load-balancer initialization until the first `dispatch()` call. Most worker subprocesses never invoke nested `@wool.routine` calls, so eager proxy startup wastes resources on discovery overhead that is never used. With lazy startup, workers only pay the cost of proxy initialization when they actually dispatch sub-tasks.

Separate `WorkerProxy`'s lifecycle into two layers: `enter()`/`exit()` manage the `wool.__proxy__` context variable, while `start()`/`stop()` manage the actual resource lifecycle (discovery, load balancer, sentinel). This split makes lazy behavior easy to reason about — `enter()` always sets the context variable, then either calls `start()` eagerly or defers it to the first `dispatch()`. A double-checked lock in `dispatch()` ensures the proxy starts exactly once under concurrent access. The `lazy` flag is preserved through `cloudpickle` serialization, so workers receiving tasks inherit the same laziness setting from the pool that dispatched them.

Closes #54

### Context management layer: enter and exit

Add `enter()` and `exit()` methods that own the `wool.__proxy__` context variable token. `enter()` unconditionally sets the context var; when `lazy=False`, it also calls `start()`. `exit()` resets the context var, then delegates to `stop()` if the proxy was started. `__aenter__`/`__aexit__` delegate to `enter()`/`exit()` respectively.

The `_proxy_token` is initialized to `None` in `__init__` and set exactly once in `enter()`, eliminating the token leak that would occur if multiple methods called `wool.__proxy__.set(self)`.

### Resource lifecycle: start and stop

`start()` (formerly `_start()`) acquires resources: subscribes to discovery, initializes the load-balancer context, and launches the worker sentinel task. `stop()` releases them. Both raise `RuntimeError` when called in an invalid state. These methods are now public, making the resource lifecycle explicit and independently callable from the context management layer.

### Lazy parameter on WorkerProxy and WorkerPool

`WorkerProxy.__init__` accepts a new `lazy` parameter (default `True`). `WorkerPool.__init__` accepts the same parameter and forwards it to every `WorkerProxy` it constructs across all pool modes (default, ephemeral, durable, durable-shared, durable-joined, hybrid). The flag flows through `__reduce__` so that pickled proxies sent to worker subprocesses preserve the laziness setting.

### Simplified proxy factory

`_proxy_factory` in `process.py` now calls `proxy.enter()` instead of `proxy.start()`, and `_proxy_finalizer` calls `proxy.exit()`. This correctly enters the context management layer, letting lazy proxies defer startup and non-lazy proxies start eagerly.

### Integration test dimension

`LazyMode` (LAZY, EAGER) is added as a ninth dimension to the pairwise scenario matrix, ensuring lazy and eager proxy behavior are exercised across all existing pool mode, discovery, credential, and timeout combinations.

## Test cases

| 1 | `TestWorkerProxy` | A discovery service | WorkerProxy is instantiated with defaults | Lazy is True | Default lazy parameter |
| 2 | `TestWorkerProxy` | A discovery service and lazy=False | WorkerProxy is instantiated | Lazy is False | Explicit eager parameter |
| 3 | `TestWorkerProxy` | An arbitrary boolean value for lazy | WorkerProxy is instantiated with that value | The lazy property matches the constructor argument | Property-based lazy round-trip |
| 4 | `TestWorkerProxy` | A lazy WorkerProxy not yet entered | `enter()` is called | Proxy remains un-started | Lazy enter defers startup |
| 5 | `TestWorkerProxy` | A non-lazy WorkerProxy not yet entered | `enter()` is called | Started is True | Eager enter starts immediately |
| 6 | `TestWorkerProxy` | A non-lazy WorkerProxy not started | `exit()` is called | Raises RuntimeError | Non-lazy exit requires prior start |
| 7 | `TestWorkerProxy` | A lazy WorkerProxy never started | `exit()` is called | Returns without raising | Graceful exit on un-started lazy proxy |
| 8 | `TestWorkerProxy` | A lazy WorkerProxy that was entered and started | `exit()` is called | Stops the proxy, started is False | Exit delegates to stop for started lazy proxy |
| 9 | `TestWorkerProxy` | A WorkerProxy with lazy=False | Pickled and unpickled | Lazy flag is preserved as False | Serialization round-trip for lazy=False |
| 10 | `TestWorkerProxy` | A WorkerProxy with default lazy=True | Pickled and unpickled | Lazy flag is preserved as True | Serialization round-trip for lazy=True |
| 11 | `TestWorkerProxy` | A lazy WorkerProxy not yet started | `dispatch()` is called | Auto-starts and dispatches the task | Lazy auto-start on first dispatch |
| 12 | `TestWorkerProxy` | A lazy WorkerProxy not yet started | Two `dispatch()` calls are made concurrently | Proxy starts once, both dispatches succeed | Double-checked locking under concurrency |
| 13 | `TestWorkerPool` | No explicit lazy parameter | WorkerPool is instantiated with spawn=2 | Constructed without error | Default lazy accepted |
| 14 | `TestWorkerPool` | Explicit lazy=False | WorkerPool is instantiated with spawn=2 | Constructed without error | Explicit lazy=False accepted |
| 15 | `TestPoolComposition` | Scenarios with LazyMode.LAZY | Pools are built and routines dispatched | Correct results returned | End-to-end lazy proxy across pool modes |
| 16 | `TestPoolComposition` | A scenario with LazyMode.EAGER | Pool is built and routine dispatched | Correct result returned | End-to-end eager proxy |
| 17 | `test_integration` | Pairwise scenarios including LazyMode dimension | Routines are dispatched | Correct results across all combinations | Pairwise coverage of lazy/eager with all other dimensions |

### DIFF
--- a/wool/src/wool/runtime/worker/README.md
+++ b/wool/src/wool/runtime/worker/README.md
@@ -193,7 +193,9 @@ Signal handlers map `SIGTERM` to timeout 0 (cancel immediately) and `SIGINT` to 
 
 ### Nested routines
 
-Worker subprocesses can dispatch tasks to other workers. Each subprocess is configured with a `ResourcePool` of `WorkerProxy` instances (via `wool.__proxy_pool__`), so `@wool.routine` calls within a task transparently route to the target pool. Spinning up a `WorkerProxy` is not free — it involves establishing a discovery subscription and opening gRPC connections — so the resource pool caches proxies with a configurable TTL (default 60 seconds, set via `proxy_pool_ttl` on `LocalWorker`). If the interval between dispatches for a given pool on a given worker is shorter than the TTL, the cached proxy is reused. If it exceeds the TTL, the proxy is finalized and must be recreated on the next dispatch. Tuning `proxy_pool_ttl` above the expected dispatch interval keeps proxies warm and avoids this cold-start overhead.
+Worker subprocesses can dispatch tasks to other workers. Each subprocess is configured with a `ResourcePool` of `WorkerProxy` instances (via `wool.__proxy_pool__`), so `@wool.routine` calls within a task transparently route to the target pool. Spinning up a `WorkerProxy` is not free — it involves establishing a discovery subscription, starting a sentinel task, and opening gRPC connections — so the resource pool caches proxies with a configurable TTL (default 60 seconds, set via `proxy_pool_ttl` on `LocalWorker`). If the interval between dispatches for a given pool on a given worker is shorter than the TTL, the cached proxy is reused. If it exceeds the TTL, the proxy is finalized and must be recreated on the next dispatch. Tuning `proxy_pool_ttl` above the expected dispatch interval keeps proxies warm and avoids this cold-start overhead.
+
+Proxies on worker subprocesses are lazy by default — the `WorkerPool` propagates its `lazy` flag to every `WorkerProxy` it constructs, and each task serializes the proxy (including the flag) so that workers receiving the task inherit the same laziness setting. A lazy proxy defers discovery subscription and sentinel setup until its first `dispatch()` call, so workers that never invoke nested routines pay no startup cost.
 
 ## Connections
 
@@ -206,6 +208,17 @@ Worker subprocesses can dispatch tasks to other workers. Each subprocess is conf
 | Pool URI | `pool_uri` | Subscribes to `LocalDiscovery` with the URI as namespace and tag filter. |
 | Discovery | `discovery` | Accepts any `DiscoverySubscriberLike` or `Factory` thereof. |
 | Static | `workers` | Takes a sequence of `WorkerMetadata` directly — no discovery needed. |
+
+### Lazy startup
+
+`WorkerProxy` accepts a `lazy` parameter (default `True`) that controls when the proxy actually starts — i.e., when it subscribes to discovery, launches the worker sentinel task, and initializes the load balancer context.
+
+| `lazy` | `start()` / `__aenter__` | `dispatch()` | `stop()` on un-started proxy |
+| ------ | ------------------------ | ------------- | ---------------------------- |
+| `True` | No-op | Starts the proxy on first call, then dispatches | No-op (safe to call) |
+| `False` | Starts eagerly | Raises `RuntimeError` if not started | Raises `RuntimeError` |
+
+When `lazy=True`, concurrent `dispatch()` calls use a double-checked lock to ensure the proxy starts exactly once. The `lazy` flag is preserved through `cloudpickle` serialization, so proxies sent to worker subprocesses as part of a task retain their laziness setting.
 
 ### Self-describing connections
 
--- a/wool/src/wool/runtime/worker/README.md
+++ b/wool/src/wool/runtime/worker/README.md
@@ -213,10 +213,10 @@ Proxies on worker subprocesses are lazy by default — the `WorkerPool` propagat
 
 `WorkerProxy` accepts a `lazy` parameter (default `True`) that controls when the proxy actually starts — i.e., when it subscribes to discovery, launches the worker sentinel task, and initializes the load balancer context.
 
-| `lazy` | `start()` / `__aenter__` | `dispatch()` | `stop()` on un-started proxy |
-| ------ | ------------------------ | ------------- | ---------------------------- |
-| `True` | No-op | Starts the proxy on first call, then dispatches | No-op (safe to call) |
-| `False` | Starts eagerly | Raises `RuntimeError` if not started | Raises `RuntimeError` |
+| `lazy` | `enter()` / `__aenter__` | `dispatch()` | `exit()` on un-started proxy |
+| ------ | ------------------------ | ------------- | ----------------------------- |
+| `True` | Sets context var only | Calls `start()` on first call, then dispatches | No-op (safe to call) |
+| `False` | Sets context var, calls `start()` | Raises `RuntimeError` if not started | Raises `RuntimeError` |
 
 When `lazy=True`, concurrent `dispatch()` calls use a double-checked lock to ensure the proxy starts exactly once. The `lazy` flag is preserved through `cloudpickle` serialization, so proxies sent to worker subprocesses as part of a task retain their laziness setting.
 
--- a/wool/src/wool/runtime/worker/proxy.py
+++ b/wool/src/wool/runtime/worker/proxy.py
@@ -35,6 +35,8 @@ from wool.runtime.worker.connection import WorkerConnection
 from wool.runtime.worker.metadata import WorkerMetadata
 
 if TYPE_CHECKING:
+    from contextvars import Token
+
     from wool.runtime.routine.task import Task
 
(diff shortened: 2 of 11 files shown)


wool/src/wool/runtime/worker/proxy.py
```diff
@@ -429,11 +453,20 @@ async def start(self) -> None:
...
         :raises RuntimeError:
-            If the proxy was not started first.
+            If the proxy was not started first and ``lazy`` is
+            ``False``.
         """
         if not self._started:
-            raise RuntimeError("Proxy not started - call start() first")
+            if not self._lazy:
+                raise RuntimeError("Proxy not started - call start() first")
+            if hasattr(self, "_proxy_token"):
```
conradbzura: Default `self._proxy_token` to None on the class.

wool/src/wool/runtime/worker/proxy.py
```diff
@@ -401,9 +413,21 @@ def workers(self) -> list[WorkerMetadata]:
...
+        :meth:`dispatch` is first called.  When ``lazy=False``, starts
+        eagerly.
+
         :raises RuntimeError:
-            If the proxy has already been started.
+            If the proxy has already been started and ``lazy`` is
+            ``False``.
         """
+        if self._lazy:
+            self._proxy_token = wool.__proxy__.set(self)
```
conradbzura: This should happen indiscriminately.

wool/src/wool/runtime/worker/proxy.py
conradbzura: This should be removed, `__proxy__` will be set in `start()`.

wool/src/wool/runtime/worker/proxy.py
```diff
@@ -259,8 +263,11 @@ def __init__(
 
         self._id: uuid.UUID = uuid.uuid4()
         self._started = False
+        self._lazy = lazy
+        self._start_lock = asyncio.Lock() if lazy else None
         self._loadbalancer = loadbalancer
         self._lease = lease
+        self._proxy_token: object | None = None
```
conradbzura: This should be annotated as `contextvars.Token | None`.

wool/src/wool/runtime/worker/README.md
```diff
@@ -193,7 +193,9 @@ Signal handlers map `SIGTERM` to timeout 0 (cancel immediately) and `SIGINT` to
...
+
+Proxies on worker subprocesses are lazy by default — the `WorkerPool` propagates its `lazy` flag to every `WorkerProxy` it constructs, and each task serializes the proxy (including the flag) so that workers receiving the task inherit the same laziness setting. A lazy proxy defers discovery subscription and sentinel setup until its first `dispatch()` call, so workers that never invoke nested routines pay no startup cost. Most tasks execute their callable and return a result without dispatching sub-tasks, making lazy the efficient default.
```
conradbzura: Drop "Most tasks execute their callable and return a result without dispatching sub-tasks, making lazy the efficient default."

wool/tests/runtime/worker/test_process.py
```diff
@@ -215,6 +215,73 @@ def mock_signal(sig, handler):
...
+        _proxy_factory() is called.
+    Then:
+        It should call start() and return the proxy.
+    """
+    # Arrange
+    mock_proxy = mocker.MagicMock()
+    mock_proxy.start = mocker.AsyncMock()
+
+    # Act
+    result = await _proxy_factory(mock_proxy)
```
conradbzura: We shouldn't test private functions directly, but rather evaluate their behavior through the public API.

wool/tests/runtime/worker/test_process.py
```diff
@@ -215,6 +215,73 @@ def mock_signal(sig, handler):
...
+        _proxy_finalizer() is called.
+    Then:
+        It should call proxy.stop().
+    """
+    # Arrange
+    mock_proxy = mocker.MagicMock()
+    mock_proxy.stop = mocker.AsyncMock()
+
+    # Act
+    await _proxy_finalizer(mock_proxy)
```
conradbzura: Same here.

wool/tests/runtime/worker/test_process.py
```diff
@@ -215,6 +215,73 @@ def mock_signal(sig, handler):
...
+        _proxy_finalizer() is called.
+    Then:
+        It should catch the exception and complete without propagating it.
+    """
+    # Arrange
+    mock_proxy = mocker.MagicMock()
+    mock_proxy.stop = mocker.AsyncMock(side_effect=Exception("Stop failed"))
+
+    # Act — should not raise exception
+    await _proxy_finalizer(mock_proxy)
```
conradbzura: And here.

wool/tests/runtime/worker/test_proxy.py
```diff
@@ -708,21 +740,41 @@ async def test_start_sets_started_flag(
...
+            It should remain un-started.
+        """
+        # Arrange
+        proxy = WorkerProxy(discovery=mock_discovery_service)
+
+        # Act
+        await proxy.start()
+
+        # Assert
+        assert not proxy.started
```
conradbzura, line 776: This is actually pretty confusing behavior. I think we need to rename `start()` and `stop()` to better reflect the logic they execute.

wool/tests/runtime/worker/test_pool.py
```diff
@@ -403,6 +403,34 @@ def test___init___durable_mode_with_discovery(self, mocker):
...
+            Explicit lazy=False.
+        When:
+            WorkerPool is instantiated with spawn=2.
+        Then:
+            It should be constructed without error.
+        """
+        # Act & Assert — no error
+        pool = WorkerPool(spawn=2, lazy=False)
+        assert isinstance(pool, WorkerPool)
+
```
conradbzura: These tests are pointless – drop them.

PR #146, opened 2026-03-31
Guard WorkerPool and WorkerProxy against reentrant context usage — Closes #145

### LINKED ISSUE #145: Guard WorkerPool and WorkerProxy against reentrant context usage
## Description

Add explicit reentry guards to `WorkerPool` and `WorkerProxy` so that calling `enter()` (or `__aenter__`) twice without an intervening `exit()` raises a clear `RuntimeError` instead of silently corrupting internal state.

Currently, calling `WorkerProxy.enter()` twice overwrites `self._proxy_token` with a new context variable token, leaking the first token. This means `exit()` can only reset the second token — the context variable retains the proxy value from the first `enter()` after the proxy is fully torn down. For non-lazy proxies the second `enter()` also calls `start()`, which raises `RuntimeError("Proxy already started")`, but the token is already overwritten by that point.

`WorkerPool.__aenter__` has a similar gap — calling it twice overwrites `self._proxy_context` with a new context manager, orphaning the first one and its associated workers.

## Motivation

The token leak is a latent correctness bug. It is invisible in tests because each test creates a fresh async context, but it would surface in long-lived applications that accidentally nest `async with proxy:` blocks or call `enter()` in a retry loop. A guard makes the mistake immediately obvious rather than silently corrupting state.

## Expected outcome

- `WorkerProxy.enter()` raises `RuntimeError` if the proxy is already entered (i.e., `_proxy_token` is not `None`).
- `WorkerPool.__aenter__` raises `RuntimeError` if the pool is already entered.
- Tests cover both the guard and the error message for each class.

conradbzura:
## Summary

Adds single-use enforcement to `WorkerPool` and `WorkerProxy` context managers so that any second entry — whether reentrant or after a full enter/exit cycle — raises `RuntimeError` immediately rather than silently corrupting state.

The implementation introduces a generic `@noreentry` descriptor class in `src/wool/utilities/noreentry.py`. The descriptor exclusively decorates instance methods (bare functions and unbound calls are static errors caught at the type-check level via `__call__ -> Never`). Guard state for each instance is tracked via a `weakref.WeakSet` on the descriptor, keeping instances clean without namespace pollution. `functools.update_wrapper` and `inspect.markcoroutinefunction` ensure the wrapper is transparent to introspection. The decorator is applied to `WorkerProxy.enter()` and `WorkerPool.__aenter__()`. Both READMEs are updated to document the single-use contract.

Closes #145.

### `src/wool/utilities/noreentry.py` — new `@noreentry` descriptor

Implements `noreentry` as a descriptor class for single-use method guards. The decorator uses the descriptor protocol's `__get__` to return a bound wrapper on instance access, and `__call__ -> Never` to reject bare function usage at the type-check level.

Guard state is tracked via a `weakref.WeakSet` on the descriptor instance — when an instance's guarded method is called, that instance is added to the WeakSet. Any subsequent call to the same method on the same instance detects the instance in the set and raises `RuntimeError`. This approach avoids polluting instance namespaces and automatically cleans up references when instances are garbage collected.

The decorator works with both sync and async methods. Attempting to use it on bare functions or call it directly (unbound style) raises `TypeError` at runtime and is flagged as a static type error via the `Never` return annotation on `__call__`.

### `src/wool/runtime/worker/proxy.py` and `pool.py` — apply guard

`@noreentry` is applied to `WorkerProxy.enter()` and `WorkerPool.__aenter__()`. The `:raises RuntimeError:` docstring on each method is updated to document single-use semantics.

### Documentation

`README.md` and `src/wool/runtime/worker/README.md` each receive a note on the single-use contract with a brief code example showing the correct pattern (create a new instance for a new context).

## Test cases

|---|---|---|---|---|---|
| 1 | `TestNoReentry` | A class with a sync `@noreentry` method | Method is called for the first time | Returns normally | First-call success for bound sync method |
| 2 | `TestNoReentry` | A class with a sync `@noreentry` method called once | Method is called a second time on the same instance | Raises `RuntimeError` | Guard on second sync call |
| 3 | `TestNoReentry` | A class with an async `@noreentry` method | Method is awaited for the first time | Returns normally | First-call success for bound async method |
| 4 | `TestNoReentry` | A class with an async `@noreentry` method awaited once | Method is awaited a second time | Raises `RuntimeError` | Guard on second async call |
| 5 | `TestNoReentry` | Two instances of a class with a `@noreentry` method, first called | Method is called on the second instance | Returns normally | Per-instance isolation |
| 6 | `TestNoReentry` | A sync `@noreentry` method called once | Method is called a second time | `RuntimeError` message includes the method's `__qualname__` | Error message content |
| 7 | `TestNoReentry` | A class with an async `@noreentry` method | `asyncio.iscoroutinefunction` is called on the method | Returns `True` | Coroutine function introspection transparency |
| 8 | `TestNoReentry` | A class with a sync `@noreentry` method | `__name__` is inspected on the decorated method | Equals the original function name | `functools.wraps` preservation |
| 9 | `TestNoReentry` | A class with two `@noreentry` methods; first called twice | Second method is called | Returns normally | Per-method dict key independence |
| 10 | `TestNoReentry` | A class with a `@noreentry` method | Method is accessed through the class (unbound) | Returns the descriptor itself | Descriptor protocol for unbound access |
| 11 | `TestNoReentry` | A `@noreentry` descriptor accessed unbound | Descriptor is called directly | Raises `TypeError` | Rejection of bare function usage |
| 12 | `TestWorkerProxy` | A `WorkerProxy` that has already been entered | `enter()` is called a second time | Raises `RuntimeError` | Reentrant entry guard on proxy |
| 13 | `TestWorkerProxy` | A `WorkerProxy` that has been entered and exited | `enter()` is called again | Raises `RuntimeError` | Post-exit re-entry guard on proxy |
| 14 | `TestWorkerPool` | A `WorkerPool` already entered via `async with` | Pool is entered again via `async with` | Raises `RuntimeError` | Reentrant entry guard on pool |
| 15 | `TestWorkerPool` | A `WorkerPool` entered and exited via `async with` | Pool is entered again via `async with` | Raises `RuntimeError` | Post-exit re-entry guard on pool |

### DIFF
--- a/wool/src/wool/utilities/noreentry.py
+++ b/wool/src/wool/utilities/noreentry.py
@@ -1,0 +1,107 @@
+from __future__ import annotations
+
+import asyncio.coroutines
+import functools
+import inspect
+import sys
+import weakref
+from typing import Never
+
+
+class _Token:
+    """Hashable, weakly-referenceable token for instance tracking."""
+
+    pass
+
+
+class NoReentryBoundMethod:
+    """Descriptor implementing single-use guard for bound methods.
+
+    On the first call the decorated method executes normally. Any
+    subsequent call raises :class:`RuntimeError`.
+
+    Guard state uses a per-instance token stored on the instance under
+    ``__noreentry_token__``. The token is unique, hashable, and tied to the
+    instance's lifetime. The descriptor tracks tokens in a WeakSet to auto-clean
+    when instances are garbage collected.
+
+    Works with both synchronous and asynchronous methods. Only supports
+    bound methods; using @noreentry on bare functions raises TypeError.
+    """
+
+    def __init__(self, fn, /):
+        functools.update_wrapper(self, fn)
+        if inspect.iscoroutinefunction(fn):
+            if sys.version_info >= (3, 12):
+                inspect.markcoroutinefunction(self)
+            else:
+                self._is_coroutine = asyncio.coroutines._is_coroutine
+        self._fn = fn
+        self._invocations = weakref.WeakSet()
+
+    def __call__(self, *args, **kwargs) -> Never:
+        raise TypeError("@noreentry only decorates methods, not bare functions")
+
+    def __get__(self, obj, objtype=None):
+        if obj is None:
+            return self
+
+        # Cache the wrapper on the instance to avoid recreating it on every access.
+        cache_key = f"__noreentry_wrapper_{id(self)}__"
+        return obj.__dict__.setdefault(cache_key, self._make_wrapper(obj))
+
+    def _make_wrapper(self, obj):
+        """Create the bound wrapper for an instance."""
+        guard = self._guard
+        fn = self._fn
+        if inspect.iscoroutinefunction(fn):
+
+            @functools.wraps(fn)
+            async def async_wrapper(*args, **kwargs):
+                guard(obj)
+                return await fn(obj, *args, **kwargs)
+
+            return async_wrapper
+        else:
+
+            @functools.wraps(fn)
+            def sync_wrapper(*args, **kwargs):
+                guard(obj)
+                return fn(obj, *args, **kwargs)
+
+            return sync_wrapper
+
+    def _guard(self, obj):
+        """Check and record invocation on the specified object."""
+        # Get or create a unique token for this object.
+        token = obj.__dict__.setdefault("__noreentry_token__", _Token())
+
+        # Check if this descriptor was invoked on this object already.
+        if token in self._invocations:
+            raise RuntimeError(
+                f"'{self._fn.__qualname__}' cannot be invoked more than once"
+            )
+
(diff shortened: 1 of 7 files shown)


wool/src/wool/utilities/noreentry.py
```diff
@@ -0,0 +1,74 @@
...
+                return fn(obj, *args, **kwargs)
+
+        return bound
+
+    def __call__(self, *args, **kwargs):
+        if self._bound:
+            # Unbound method call: SomeClass.method(instance, ...)
+            self._guard(args[0])
+            return self._fn(*args, **kwargs)
+        if self._invoked:
```
conradbzura: Can't we just use this path for all decorated function types, including bound methods?

PR #149, opened 2026-04-01
Add admission control hook for worker backpressure — Closes #57

### LINKED ISSUE #57: Add admission control hook for worker backpressure
## Description

Add a user-configurable admission control hook to workers that determines whether an incoming task should be accepted or rejected. The hook is a callable that receives context about the worker's current state (e.g., active task count, task metadata) and returns an accept/reject decision. When a task is rejected, the worker responds with gRPC `RESOURCE_EXHAUSTED`, which the load balancer already classifies as transient — causing it to skip to the next worker automatically.

The default behavior (no hook configured) accepts all tasks, preserving the current semantics. The existing `WorkerConnection` semaphore continues to cap per-worker concurrent dispatches on the client side; this hook adds server-side admission control.

## Motivation

Workers currently have no way to signal that they are overloaded. The client-side semaphore limits concurrency per connection, but the worker itself accepts all incoming tasks unconditionally. In production, workers may need to reject tasks based on workload-specific criteria — active task count, memory pressure, GPU utilization, or custom application logic.

Backpressure is inherently workload-specific, so a hook-based approach lets users define the rejection predicate without Wool prescribing what "overloaded" means. The gRPC `RESOURCE_EXHAUSTED` → transient error → skip-to-next-worker path is already fully implemented in the load balancer, so the only missing piece is letting the worker trigger it.

## Expected outcome

- `LocalWorker` accepts an optional `backpressure` parameter — a callable that receives worker state and returns whether to reject the task.
- A rejected task triggers gRPC `RESOURCE_EXHAUSTED`, causing the load balancer to skip to the next available worker.
- The hook is invoked after task deserialization but before acknowledgment, so rejected tasks never enter the worker's active task set.
- Both sync and async hook implementations are supported.
- When no hook is configured, all tasks are accepted (backward compatible).

conradbzura:
## Summary

Add a `BackpressureLike` protocol and `BackpressureContext` dataclass that let users configure a per-worker hook to reject incoming tasks based on workload-specific criteria. The hook receives a snapshot of the worker's active task count and the incoming task, returning `True` to reject (triggering gRPC `RESOURCE_EXHAUSTED`) or `False` to accept. The load balancer already treats `RESOURCE_EXHAUSTED` as transient and skips to the next worker — so no client-side changes are needed.

The default behavior (no hook) accepts all tasks, preserving current semantics. The existing `WorkerConnection` semaphore continues to cap per-worker concurrent dispatches on the client side; this hook adds server-side admission control.

Closes #57

### BackpressureLike protocol and BackpressureContext dataclass (`service.py`)

Introduce two new public types in `wool.runtime.worker.service`:

- `BackpressureContext(frozen=True)` — immutable snapshot with `active_task_count: int` and `task: Task`
- `BackpressureLike(Protocol)` — runtime-checkable protocol accepting both sync and async callables. Return `True` to reject, `False` to accept. The falsey default of `None` (no hook) naturally means "accept all".

### Backpressure check in WorkerService.dispatch (`service.py`)

Insert the check after task deserialization but before the Ack response. If the hook returns truthy, the service calls `context.abort(StatusCode.RESOURCE_EXHAUSTED, ...)` — the task never enters the docket and no Ack is sent. Both sync and async hooks are supported via `inspect.isawaitable()`.

### Thread hook through WorkerProcess (`process.py`)

Accept `backpressure` in `WorkerProcess.__init__` and serialize it with `cloudpickle.dumps()` to cross the `multiprocessing.spawn` boundary. This supports lambdas, closures, and callable class instances — not just top-level functions. Deserialize with `cloudpickle.loads()` in `_serve()` before passing to `WorkerService`.

### User-facing API on LocalWorker (`local.py`)

Add `backpressure: BackpressureLike | None = None` parameter to `LocalWorker.__init__`, passed through to `WorkerProcess`. Usage with `WorkerPool` is supported via `functools.partial`:

```python
from functools import partial

worker_factory = partial(LocalWorker, backpressure=my_hook)
async with WorkerPool(worker=worker_factory):
    ...
```

### Public exports (`__init__.py`)

Export `BackpressureContext` and `BackpressureLike` from the top-level `wool` package.

## Test cases

| 1 | `TestBackpressureContext` | Valid active_task_count and Task | Instantiated | Stores both fields correctly | Field storage |
| 2 | `TestBackpressureContext` | A BackpressureContext instance | Attribute reassigned | Raises AttributeError | Frozen immutability |
| 3 | `TestBackpressureContext` | Two instances with identical fields | Compared with `==` | Returns True | Equality |
| 4 | `TestBackpressureContext` | Two instances with different count | Compared with `==` | Returns False | Inequality |
| 5 | `TestBackpressureContext` | Any non-negative int (Hypothesis) | Instantiated | Field equals input | Arbitrary task count |
| 6 | `TestBackpressureLike` | Sync function | `isinstance` check | Returns True | Sync protocol |
| 7 | `TestBackpressureLike` | Async function | `isinstance` check | Returns True | Async protocol |
| 8 | `TestBackpressureLike` | Callable class instance | `isinstance` check | Returns True | Class protocol |
| 9 | `TestBackpressureLike` | A plain string | `isinstance` check | Returns False | Negative case |
| 10 | `TestWorkerService` | Callable hook | `WorkerService(backpressure=hook)` | Initializes with events unset | Init with hook |
| 11 | `TestWorkerService` | Sync hook returning False | Dispatch RPC called | Accepts task, returns result | Sync accept |
| 12 | `TestWorkerService` | Sync hook returning True | Dispatch RPC called | RESOURCE_EXHAUSTED raised | Sync reject |
| 13 | `TestWorkerService` | Async hook returning True | Dispatch RPC called | RESOURCE_EXHAUSTED raised | Async reject |
| 14 | `TestWorkerService` | Async hook returning False | Dispatch RPC called | Accepts task, returns result | Async accept |
| 15 | `TestWorkerService` | Hook capturing argument | Dispatch RPC called | Receives BackpressureContext with correct fields | Context validation |
| 16 | `TestWorkerService` | One active task dispatched | Second dispatch with hook | Hook sees active_task_count == 1 | Active count |
| 17 | `TestWorkerService` | Hook that raises RuntimeError | Dispatch RPC called | Propagates as gRPC failure | Hook exception |
| 18 | `TestWorkerProcess` | Callable hook | `WorkerProcess(backpressure=hook)` | Constructs without error | Init with hook |
| 19 | `TestLocalWorker` | Callable hook | `LocalWorker(backpressure=hook)` | Forwards hook to WorkerProcess | Init with hook |
| 20 | `TestBuildPoolFromScenario` | Sync backpressure hook | Pool built and routine dispatched | Returns correct result | Sync hook integration |
| 21 | `TestBuildPoolFromScenario` | Async backpressure hook | Pool built and routine dispatched | Returns correct result | Async hook integration |
| 22 | `TestBackpressureRejection` | Sync reject hook, single worker | Routine dispatched | Raises NoWorkersAvailable | Sync rejection e2e |
| 23 | `TestBackpressureRejection` | Async reject hook, single worker | Routine dispatched | Raises NoWorkersAvailable | Async rejection e2e |
| 24 | `TestBackpressureRejection` | Two workers, one rejecting | Routine dispatched | Falls through to accepting worker | Load balancer fallback |

### DIFF
--- a/wool/src/wool/runtime/worker/local.py
+++ b/wool/src/wool/runtime/worker/local.py
@@ -64,6 +64,15 @@ class LocalWorker(Worker):
     :param options:
         gRPC message size options. Defaults to
         :class:`WorkerOptions` with 100 MB limits.
+    :param backpressure:
+        Optional admission control hook. A callable receiving a
+        :class:`~wool.runtime.worker.service.BackpressureContext`
+        and returning ``True`` to **reject** the task or ``False``
+        to **accept** it. Both sync and async callables are
+        supported. When a task is rejected the worker responds with
+        gRPC ``RESOURCE_EXHAUSTED``, causing the load balancer to
+        skip to the next worker. ``None`` (default) accepts all
+        tasks unconditionally.
     :param extra:
         Additional metadata as key-value pairs.
     """
--- a/wool/src/wool/runtime/worker/local.py
+++ b/wool/src/wool/runtime/worker/local.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from typing import TYPE_CHECKING
 from typing import Any
 
 import grpc.aio
@@ -10,6 +11,9 @@ from wool.runtime.worker.auth import WorkerCredentials
 from wool.runtime.worker.base import Worker
 from wool.runtime.worker.base import WorkerOptions
 from wool.runtime.worker.process import WorkerProcess
+
+if TYPE_CHECKING:
+    from wool.runtime.worker.service import BackpressureLike
 
 
 # public
@@ -76,6 +80,7 @@ class LocalWorker(Worker):
         proxy_pool_ttl: float = 60.0,
         credentials: WorkerCredentials | None = None,
         options: WorkerOptions | None = None,
+        backpressure: BackpressureLike | None = None,
         **extra: Any,
     ):
         super().__init__(*tags, **extra)
@@ -90,6 +95,7 @@ class LocalWorker(Worker):
             options=options,
             tags=frozenset(self._tags),
             extra=self._extra,
+            backpressure=backpressure,
         )
 
     @property
--- a/wool/src/wool/__init__.py
+++ b/wool/src/wool/__init__.py
@@ -38,6 +38,8 @@ from wool.runtime.worker.local import LocalWorker
 from wool.runtime.worker.metadata import WorkerMetadata
 from wool.runtime.worker.pool import WorkerPool
 from wool.runtime.worker.proxy import WorkerProxy
+from wool.runtime.worker.service import BackpressureContext
+from wool.runtime.worker.service import BackpressureLike
 from wool.runtime.worker.service import WorkerService
 
 pickling_support.install()
@@ -80,6 +82,9 @@ __all__ = [
     "TaskException",
     "current_task",
     "routine",
+    # Backpressure
+    "BackpressureContext",
+    "BackpressureLike",
     # Workers
     "LocalWorker",
     "Worker",
--- a/wool/src/wool/runtime/worker/process.py
+++ b/wool/src/wool/runtime/worker/process.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import Final
 
+import cloudpickle
 import grpc.aio
 
 import wool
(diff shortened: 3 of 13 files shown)


wool/src/wool/runtime/worker/local.py
```diff
@@ -76,6 +80,7 @@ def __init__(
         proxy_pool_ttl: float = 60.0,
         credentials: WorkerCredentials | None = None,
         options: WorkerOptions | None = None,
+        backpressure: BackpressureLike | None = None,
```
conradbzura, line 92: Need to document this in the docstring. Summarize the details of `BackpressureLike`.

PR #156, opened 2026-04-06
Refactor load balancer protocol to delegate worker selection instead of dispatching tasks — Closes #155

### LINKED ISSUE #155: Refactor load balancer protocol to delegate worker selection instead of dispatching tasks
## Description

Refactor the load balancer protocol so that the load balancer's responsibility is limited to **worker selection and routing policy**. Instead of receiving a `Task` and returning an `AsyncGenerator` of results (which requires the load balancer to call `connection.dispatch` internally), the new protocol defines a `delegate` method that returns a bidirectional async generator yielding `(WorkerMetadata, WorkerConnection)` pairs. The proxy communicates dispatch outcomes back to the generator via `asend()` on success and `athrow()` on failure.

The proxy becomes responsible for dispatching the task over the connection and for evicting workers on non-transient errors. The load balancer receives full visibility into dispatch outcomes — success confirmations and exceptions — but operates on a read-only view of the worker context and cannot mutate pool membership.

### New protocol

```python
@runtime_checkable
class DelegatingLoadBalancerLike(Protocol):
    async def delegate(
        self, *, context: LoadBalancerContextLike,
    ) -> AsyncGenerator[tuple[WorkerMetadata, WorkerConnection], WorkerMetadata | None]: ...
```

The generator supports three signals:

1. `anext()` — request a worker candidate.
2. `athrow(exc)` — notify the load balancer that the last candidate failed. On transient errors (e.g., `TransientRpcError`), the proxy skips the worker and the load balancer yields the next candidate. On non-transient errors, the proxy evicts the worker from the context before throwing, and the load balancer can react to the capacity change (e.g., spawn a replacement) before yielding the next candidate.
3. `asend(metadata)` — notify the load balancer that the dispatch to the given worker succeeded, enabling internal bookkeeping (sticky routing, in-flight count tracking, affinity).

### Worker context

The `LoadBalancerContextLike` passed to `delegate` exposes only the read-only `workers` property — no `add_worker`, `remove_worker`, or `update_worker`. The load balancer can observe pool membership but cannot mutate it. Eviction is the proxy's responsibility. This is enforced at the protocol level.

### Proxy responsibilities

`WorkerProxy.dispatch` owns the dispatch-retry-evict loop:

- Call `connection.dispatch(task, timeout=...)` on the yielded worker.
- On `TransientRpcError`: `athrow()` into the generator, receive the next candidate.
- On non-transient exception: evict the worker from the context, then `athrow()` into the generator so the load balancer can react to the capacity change and yield the next candidate.
- On success: `asend(metadata)` into the generator to confirm the dispatch.

### Backwards compatibility

The existing `LoadBalancerLike` protocol (with `dispatch`) is renamed to `DispatchingLoadBalancerLike` and deprecated. A union type alias provides the transition:

```python
# Transition alias — accepts both old and new protocols.
# Will become just DelegatingLoadBalancerLike in the next major release.
LoadBalancerLike = DispatchingLoadBalancerLike | DelegatingLoadBalancerLike
```

`WorkerProxy.dispatch` checks which protocol the load balancer conforms to at dispatch time:
- If `DelegatingLoadBalancerLike`: the proxy owns the dispatch-retry-evict loop as described above.
- If `DispatchingLoadBalancerLike` (deprecated): the proxy falls back to the current behavior, calling `loadbalancer.dispatch(task, context=..., timeout=...)` directly with a mutable context.

A deprecation warning is emitted when a `DispatchingLoadBalancerLike` is passed to `WorkerPool` or `WorkerProxy`. Support for `DispatchingLoadBalancerLike` will be dropped in the next major release.

## Motivation

The current protocol couples worker selection with task dispatch. Every custom `LoadBalancerLike` implementation must handle `connection.dispatch()`, manage the result stream lifecycle, implement error-handling retry loops, and perform worker eviction — all of which are identical across implementations. The only part that actually varies is the selection logic.

This coupling creates three problems:

1. **High implementation surface.** Custom load balancer implementations require extensive stream forwarding boilerplate. A least-connections balancer needs ~60 lines, most of which is an async generator wrapper to track in-flight counts across the stream lifecycle. The actual selection logic is a handful of lines.

2. **Subtle correctness bugs.** The stream-forwarding pattern requires careful lifecycle management. When load balancers compose (e.g., a scaling wrapper around a routing balancer), in-flight counters at different levels can desynchronize — the inner balancer's counter decrements before the outer wrapper's stream is fully consumed, leading to premature worker eviction while gRPC streams are still active. If dispatch responsibility lived in the proxy, custom balancers wouldn't encounter this class of bug.

3. **Inconsistent error semantics.** Each load balancer reimplements the same transient-vs-non-transient error handling and eviction logic. `RoundRobinLoadBalancer` checkpoints and cycles; a least-connections balancer iterates a ranked list. Both do the same thing — skip on `TransientRpcError`, evict on anything else — but each has to code it from scratch. Moving dispatch and eviction to the proxy centralizes this while still notifying the load balancer of outcomes.

## Expected outcome

- `DelegatingLoadBalancerLike` defines a `delegate` method that yields worker candidates and receives dispatch outcomes (success via `asend`, failure via `athrow`), with no knowledge of tasks, dispatch, or eviction.
- `LoadBalancerContextLike` for delegating load balancers is read-only — `workers` property only, no mutation methods.
- `DispatchingLoadBalancerLike` (the current `LoadBalancerLike` with `dispatch`) is deprecated and supported via a runtime check in the proxy.
- `LoadBalancerLike` becomes a type alias for the union, narrowing to `DelegatingLoadBalancerLike` only in the next major release.
- `WorkerProxy.dispatch` owns the dispatch-retry-evict loop for delegating load balancers.
- `RoundRobinLoadBalancer` migrates to `delegate` — its implementation reduces to cycling logic with no stream management or eviction.
- Custom load balancer implementations only need to implement worker ordering and routing policy. They receive full visibility into dispatch outcomes for internal bookkeeping but cannot mutate pool membership.

conradbzura:
## Summary

Split the load balancer protocol so that balancers own only **worker selection and routing policy**, while `WorkerProxy` owns the **dispatch-retry-evict loop**. A load balancer implements `LoadBalancerLike` — a `@runtime_checkable` protocol whose single `delegate(task, *, context)` method is an async generator that yields `(WorkerMetadata, WorkerConnection)` candidates and receives dispatch outcomes back from the proxy via `asend` (success) and `athrow` (failure). The proxy calls `WorkerConnection.dispatch` directly, classifies errors as transient vs non-transient `RpcError`, evicts workers on non-transient failures, and enforces that `asend` is terminal — yielding again after a success signal raises `RuntimeError`.

`delegate` receives the `Task` as a read-only first argument, so a balancer can key routing decisions on it — consistent/rendezvous hashing by `task.tag`, session or affinity stickiness, data/model locality, or task-type specialization — or ignore it entirely, as `RoundRobinLoadBalancer` does. Selection depends on the task; dispatch, retry, and eviction stay in the proxy.

Rename the previous dispatch-owning protocol to `DispatchingLoadBalancerLike` and deprecate it with `warnings.deprecated` (applied on Python 3.13+, where it does not perturb `@runtime_checkable`) plus a runtime `DeprecationWarning` emitted from `WorkerProxy.__init__` so the warning resolves to the caller's construction frame, with `start()` as a fallback for factory-resolved balancers. `LoadBalancerLike` is the surviving name for the go-forward delegating protocol, and the proxy accepts either protocol during the deprecation window — resolving which one **once** in `start()` so `dispatch()` carries no per-call structural protocol check.

Introduce a read-only `LoadBalancerContextView` protocol exposing only the `workers` mapping. Delegating balancers are typed against this read-only view; eviction on failure is the proxy's responsibility. The proxy passes the concrete `LoadBalancerContext`, whose `workers` mapping is runtime-immutable via `MappingProxyType`.

Migrate `RoundRobinLoadBalancer` to a pure cycling generator: no dispatch, error-classification, or eviction code, and **no lock**. Worker selection is a synchronous index advance with no `await` between the read and write of the shared per-context index, so the single-threaded event loop already serializes concurrent selections and hands each dispatch a distinct starting worker (anti-thundering-herd) without a lock. Per-context rotation state lives in a `weakref.WeakKeyDictionary`, so a retired pool's cursor and the connections it references are reclaimed with the context rather than pinned for the balancer's lifetime. Preserve the pre-refactor cancellation semantics: `CancelledError` and other `BaseException` subclasses propagate through the proxy without eviction or retry, and a stream-ownership handoff via a sentinel variable releases the gRPC call if the proxy unwinds between `connection.dispatch` returning and the stream reaching the caller.

Closes #155

### New protocols and read-only context (`loadbalancer/base.py`)

Define three public protocols:

- `LoadBalancerContextView` — `@runtime_checkable` protocol exposing only the read-only `workers` property. Passed to `LoadBalancerLike.delegate` so balancers observe pool membership without mutating it. Document a package-wide convention that a `*View` suffix denotes a read-only protocol view.
- `LoadBalancerLike` — `@runtime_checkable` protocol with a single `delegate(self, task, *, context)` async-generator method returning `AsyncGenerator[tuple[WorkerMetadata, WorkerConnection], WorkerMetadata | None]`. The generator supports three proxy-driven signals: `anext()` (next candidate), `athrow(exc)` (dispatch failed), and `asend(...)` (dispatch succeeded — the generator must terminate). The `task` argument is read-only and may be used for routing or ignored.
- `DispatchingLoadBalancerLike` — the former `LoadBalancerLike`, renamed and marked deprecated. Retains the `dispatch(task, *, context, timeout)` signature for backwards compatibility, earmarked alongside `LoadBalancerContextLike` for co-removal when the deprecation window closes.

`LoadBalancerContextLike` inherits from `LoadBalancerContextView`, preserving the mutable surface used by the proxy and the deprecated dispatch path. `NoWorkersAvailable` carries a default exhaustion message so the proxy's two raise sites are single-sourced. Hoist the deprecation message into a single module-level constant shared by the `warnings.deprecated` decorator and the proxy's runtime warning.

### RoundRobinLoadBalancer migration (`loadbalancer/roundrobin.py`)

Replace `dispatch` with `delegate(task, *, context)`, ignoring `task` (round-robin is task-agnostic). The generator:

- Holds **no lock**. It reads the shared per-context index and advances it in a single synchronous step; because nothing `await`s between the read and the write, concurrent dispatches on one context cannot interleave and each observes a distinct starting worker.
- Keeps per-context rotation state in a `weakref.WeakKeyDictionary`, so entries auto-evict when the proxy drops a context.
- Snapshots `list(context.workers.items())` once per wrap and indexes it O(1) with wraparound, binding `context.workers` once per iteration, for O(1)-per-candidate selection amortized over a cycle.
- Detects exhaustion with a checkpoint keyed on the first candidate's UID, reseeded to the next survivor when that worker is evicted mid-cycle, so a full cycle without a successful dispatch terminates and the proxy raises `NoWorkersAvailable`.
- Skips a candidate already removed from the live pool, advances on `athrow`, and terminates on the success resume.

The class docstring keeps the observable contract up front and moves the checkpoint-reseed and `WeakKeyDictionary`-reclamation mechanics into an `Implementation notes` rubric. `__reduce__` resets runtime state so an unpickled instance cycles from position zero.

### Proxy-owned dispatch loop (`worker/proxy.py`)

`WorkerProxy` resolves the balancer's protocol **once** in `start()` (binding a `self._delegating` flag) and `dispatch()` branches on that flag — no per-call `@runtime_checkable isinstance`:

- **`LoadBalancerLike`**: route through `_delegate_dispatch`, which owns the retry-evict loop. It pulls candidates from `delegate(task, context=...)`, calls `WorkerConnection.dispatch` on each, and reports the outcome back via `asend` (success) or `athrow` (failure). A single narrow `except RpcError` evicts only when the error is not a `TransientRpcError`, before notifying the balancer — so non-worker-health exceptions (`ValueError`, `ChainSerializationError`) and `BaseException` (cancellation) propagate unretried. An inner `try`/`finally` with a `stream = None` sentinel `aclose`s the gRPC stream if the success path unwinds before handoff.
- **`DispatchingLoadBalancerLike`**: fall through to the legacy `dispatch(task, context=..., timeout=...)` call.

The deprecation `DeprecationWarning` for a dispatching balancer is emitted from `WorkerProxy.__init__` (caller's construction frame), with `start()` as a fallback for factory-resolved balancers; both sites share the single deprecation-message constant. When a class implements both protocols during migration, `start()` classifies it as delegating so the new path wins, and the warning is gated on the resolved path so a dual-protocol balancer routes through `delegate` without emitting a spurious `DeprecationWarning`.

### Public exports (`wool/__init__.py`)

Export `LoadBalancerLike`, `DispatchingLoadBalancerLike`, and `LoadBalancerContextView` (alongside the existing `LoadBalancerContextLike`, `NoWorkersAvailable`, and `RoundRobinLoadBalancer`).

### Documentation

- Rewrite `loadbalancer/README.md` around the delegate protocol: describe the current two-protocol state and migration, thin the contract section to point at the `LoadBalancerLike` docstring, state the read-only view as a type-level guarantee, and add a `StickyTagBalancer` example that hashes `task.tag` onto a worker ring to demonstrate task-keyed affinity routing.
- Update `WorkerProxy`/`WorkerPool` docstring examples to the `delegate(task, *, context)` shape.
- Update `worker/README.md` error-classification table and fix the stale `discovery/README.md` parameter reference.

## Test cases

| 1 | `TestLoadBalancerContextView` | A class exposing only a read-only `workers` property | Checked against the protocol | It satisfies `LoadBalancerContextView` | Read-only context protocol |
| 2 | `TestLoadBalancerLike` | A class implementing `delegate` | Checked against the protocol | It satisfies `LoadBalancerLike`, while a dispatch-only class does not | Delegating protocol conformance |
| 3 | `TestLoadBalancerLike` | A class implementing both `delegate` and `dispatch` | Checked against both protocols | It satisfies `LoadBalancerLike` and `DispatchingLoadBalancerLike` | Dual-implementation conformance |
| 4 | `TestDispatchingLoadBalancerLike` | A class implementing `dispatch` | Checked against the protocol | It satisfies the deprecated `DispatchingLoadBalancerLike` | Deprecated protocol conformance |
| 5 | `TestLoadBalancerContext` | A context with registered workers | Workers are added, updated, upserted, or removed | The `workers` mapping reflects the change and stays a `MappingProxyType` | Context membership and immutability |
| 6 | `TestRoundRobinLoadBalancer` | A balancer with a single worker | `delegate` is driven with `anext` | The first candidate is that worker | Baseline selection |
| 7 | `TestRoundRobinLoadBalancer` | Multiple workers, one prior success or failure | `delegate` is driven again | The index advances past the previous worker | Round-robin rotation |
| 8 | `TestRoundRobinLoadBalancer` | A balancer whose workers all fail | `athrow` repeats until exhaustion | Exactly N candidates yield before `StopAsyncIteration` | Full-cycle termination |
| 9 | `TestRoundRobinLoadBalancer` | A candidate has just been yielded | The proxy reports success via `asend` | The generator terminates | Success-is-terminal contract |
| 10 | `TestRoundRobinLoadBalancer` | N workers and K successive dispatches | Each `delegate` call is driven to success | Dispatch k starts at `workers[k % N]`, including wraparound | Cross-call rotation (property) |
| 11 | `TestRoundRobinLoadBalancer` | Four concurrent drivers over four workers | Each is driven to its first candidate concurrently | Each lands on a distinct worker | Anti-thundering-herd without a lock |
| 12 | `TestRoundRobinLoadBalancer` | The checkpoint worker is evicted while survivors fail transiently forever | The generator is driven with repeated `athrow` | It terminates via `StopAsyncIteration` rather than spinning | Checkpoint reseed on eviction |
| 13 | `TestRoundRobinLoadBalancer` | An ahead-of-cursor worker is evicted after the snapshot | The generator is driven to exhaustion | The evicted worker is never yielded | Live-skip of evicted candidates |
| 14 | `TestRoundRobinLoadBalancer` | A 16-worker pool observed through a counting context view | One full cycle is driven | The worker order is materialized at most once per wrap | O(1)-per-candidate selection |
| 15 | `TestRoundRobinLoadBalancer` | A served context the caller then drops | Garbage collection is forced | A weakref to the context resolves to `None` | No per-context leak |
| 16 | `TestRoundRobinLoadBalancer` | A balancer driven through one cycle | Pickled and unpickled | The restored instance cycles from position zero | Serialization resets runtime state |
| 17 | `TestWorkerProxyDispatchRetryEviction` | A delegating balancer that records the task its `delegate` receives | `dispatch(task)` runs | The recorded task is the object passed to `dispatch()` | Task forwarded to delegate |
| 18 | `TestWorkerProxyDispatchRetryEviction` | A worker returning a transient then a non-transient `RpcError` | `dispatch()` runs | The transient is retried without eviction; the non-transient evicts before the next candidate | Error classification |
| 19 | `TestWorkerProxyDispatchRetryEviction` | A dispatch raising a non-`RpcError` exception | `dispatch()` runs | The exception propagates without eviction or retry | RpcError-contract boundary |
| 20 | `TestWorkerProxyDispatchRetryEviction` | A candidate stream is handed off, or a yield follows a success signal | `dispatch()` runs to success, or the generator yields post-`asend` | The stream reaches the caller; a post-success yield raises `RuntimeError` | Success handoff and asend-terminal |
| 21 | `TestWorkerProxyDispatchRetryEviction` | Cancellation arrives during `connection.dispatch` or during `asend` | The dispatch task is cancelled | `CancelledError` propagates and any orphaned stream is `aclose`d | Cancellation and stream ownership |
| 22 | `TestWorkerProxyDispatchRetryEviction` | Workers with a drawn per-worker outcome sequence | `dispatch()` is driven over the sequence | Surviving set equals input minus non-transient failures; first success returns else `NoWorkersAvailable`; transient failures do not evict | Retry/evict invariant (property) |
| 23 | `TestWorkerProxyDispatchRetryEviction` | A real round-robin balancer whose checkpoint fails non-transiently and whose survivors fail transiently forever | `dispatch()` is awaited under a short timeout | It raises `NoWorkersAvailable` rather than hang | Proxy-driven liveness under eviction |
| 24 | `TestWorkerProxyDispatchRetryEviction` | A delegating balancer | The proxy is started | No `DeprecationWarning` is emitted | Negative deprecation coverage |
| 25 | `TestWorkerProxyDispatchingBalancerBackcompat` | A legacy dispatching balancer passed as an instance | The proxy is constructed then started | Exactly one `DeprecationWarning` fires, at construction, and start does not warn again | Deprecation construction-frame and fire-once |
| 26 | `TestWorkerProxyDispatchingBalancerBackcompat` | A factory returning a legacy balancer | The proxy is constructed then started | No warning at construction; one `DeprecationWarning` fires when `start()` resolves the factory | Deprecation factory fallback |
| 27 | `TestWorkerProxyDispatchingBalancerBackcompat` | A legacy dispatch-only balancer | `dispatch()` runs | The legacy `dispatch` stream is returned to the caller | Backwards-compatible dispatch path |
| 28 | `TestWorkerProxyDispatchingBalancerBackcompat` | A balancer implementing both `delegate` and `dispatch` | `dispatch()` runs | It routes through `delegate` (the delegate result, not the dispatch marker) and emits no `DeprecationWarning` | Dual-protocol prefers delegate |
| 29 | `test_public` | The `wool` package public API | `__all__` is validated | It exposes `DispatchingLoadBalancerLike` and `LoadBalancerContextView` alongside the existing `LoadBalancerLike` and `LoadBalancerContextLike` | Public surface |


wool/src/wool/runtime/loadbalancer/base.py
```diff
@@ -88,17 +107,19 @@ def remove_worker(self, metadata: WorkerMetadata) -> None:
...
-
-    Load balancers implementing this protocol operate on a
-    :py:class:`LoadBalancerContextLike` to access workers and their
-    connections. The context provides isolation, allowing a single load
-    balancer instance to service multiple worker pools with independent
-    state.
-
-    The dispatch method accepts a :py:class:`Task` and returns an async
-    iterator that yields task results from the worker.
+class DispatchingLoadBalancerLike(Protocol):
```
conradbzura, line 122: We can't rename this as it's not backwards compatible.

wool/src/wool/runtime/loadbalancer/base.py
```diff
@@ -110,6 +131,78 @@ async def dispatch(
...
+    """
+
+    def delegate(
+        self,
+        *,
+        context: LoadBalancerContextView,
+    ) -> AsyncGenerator[
+        tuple[WorkerMetadata, WorkerConnection],
+        WorkerMetadata | None,
+    ]: ...
```
conradbzura: Add a docstring to `delegate`.

PR #170, opened 2026-04-07
Add quorum parameter to WorkerPool and WorkerProxy — Closes #144

### LINKED ISSUE #144: Add quorum parameter to WorkerPool and WorkerProxy
## Summary

Add a `quorum: int | None = None` parameter to `WorkerProxy` and `WorkerPool` that makes `_start()` block until `len(workers) >= quorum` before returning. This ensures a minimum number of workers are discovered before the proxy considers itself started, whether that happens eagerly via `start()` or lazily on first `dispatch()`.

## Motivation

Currently there is no way to require a minimum number of workers before dispatch begins. `_await_workers()` only waits for at least one worker. For workloads that require a minimum level of parallelism — such as map-reduce, ensemble inference, or sharded computation — callers need a guarantee that enough workers are present before the first task is routed.

## Expected outcome

- `WorkerProxy(discovery=..., quorum=3)` makes `_start()` block until at least 3 workers are discovered
- If `lease` is set and `quorum > lease`, raise `ValueError` at construction — the quorum can never be satisfied
- Lazy semantics are preserved: for `lazy=True` pools, the quorum blocks on first `dispatch()`; for `lazy=False` pools, it blocks on `start()` / `__aenter__`
- `WorkerPool` forwards `quorum` to the underlying `WorkerProxy`

conradbzura:
## Summary

Add a `quorum: int | None = 1` parameter to `WorkerProxy` and `WorkerPool` that gates dispatch on a minimum number of workers being admitted, and a sibling `quorum_timeout: float | None = 60` controlling how long to wait. The default `quorum=1` preserves the pre-PR implicit-wait semantics (every `dispatch()` used to block up to 60s for the first worker); pass `quorum=None` or `quorum=0` to disable the gate entirely.

Beyond the user-facing parameter, the proxy's start/stop/dispatch lifecycle is hardened: `start()` uses `AsyncExitStack` to atomically unwind partial setup on failure; `stop()` follows the same pattern with guaranteed teardown ordering (sentinel → discovery → loadbalancer); `enter()` resets the `wool.__proxy__` contextvar token if `start()` raises; `_teardown_sentinel` nulls every attribute populated by `start()` so a failed proxy leaves no stale references; and lazy first-dispatch failures are cached so subsequent dispatches fail-fast instead of paying `quorum_timeout` again. The cloudpickle round-trip carries `quorum` and `quorum_timeout` through `__wool_reduce__`; older reduce tuples without these fields restore via `_restore_proxy` defaults of `(1, 60)`, preserving pre-quorum semantics for cross-version peers.

Closes #144

### Quorum parameters on `WorkerProxy`

Add `quorum: int | None = 1` and `quorum_timeout: float | None = 60` across the six `__init__` overloads (three shapes — pool URI, discovery, static workers — × two quorum-or-not variants). The no-quorum overloads expose only `quorum: None`; the with-quorum overloads expose `quorum: int = 1` and `quorum_timeout`. The runtime accepts `quorum_timeout` as `Undefined` internally to distinguish user-supplied from default. Validate at construction: `quorum < 0` raises; `quorum > lease` raises; for static `workers=[...]`, `quorum > len(compatible_workers)` raises with the pre-filter count in the message (e.g., `"1 of 2 after security/version filtering"`); `quorum_timeout <= 0` raises; `quorum_timeout` supplied with `quorum in (None, 0)` raises.

### Lifecycle hardening on `WorkerProxy`

Rewrite `start()` with `contextlib.AsyncExitStack` so partial setup is unwound atomically on failure (loadbalancer entered → discovery entered → sentinel started → quorum wait; reverse order on rollback). Rewrite `stop()` with the same pattern so all three teardowns run even if one raises. Wrap `start()` in `try/except` inside `enter()` so the `_proxy_token` contextvar is reset on failure — Python doesn't invoke `__aexit__` when `__aenter__` raises. Extend `_teardown_sentinel` (the unified rollback callback) to null `_loadbalancer_context`, `_loadbalancer_service`, `_loadbalancer_context_manager`, `_discovery_stream`, `_discovery_context_manager`, and `_workers_changed` so a failed proxy carries no stale references.

`_await_workers` becomes a pure waiter on a new `_workers_changed: asyncio.Event` set by the sentinel after each `worker-added` or `worker-dropped`; `start()` is the sole gate (`if self._quorum: ...`). `dispatch()` no longer runs an unconditional 60s wait — lazy first-dispatch invokes `start()` under `_start_lock`, and if `start()` raises, the exception is cached on `_start_failure` so subsequent dispatches re-raise the same exception immediately without re-running `start()` (preventing N concurrent dispatchers from each paying `quorum_timeout`).

### Cloudpickle round-trip

`__wool_reduce__` carries `_quorum` and `_quorum_timeout` in the reduce tuple. `_restore_proxy` accepts both with back-compat defaults of `quorum=1, quorum_timeout=60` so pre-quorum reduce tuples emitted by older peers restore with the prior implicit-wait behaviour. The restoration uses typed two-branch narrowing — pass literal `None` in the no-quorum path, pass both values in the quorum path — so pyright matches `WorkerProxy`'s typed overloads cleanly.

### Quorum parameters on `WorkerPool`

Add `quorum: int | None = 1` and `quorum_timeout: float | None = 60` across ten `__init__` overloads (five pool shapes — spawn-only, discovery-only, hybrid, plus two deprecated `size=` variants — × two quorum-or-not variants). The pool's `_make_proxy` helper forwards both to the underlying `WorkerProxy` across all four runtime modes (spawn-only, discovery-only, hybrid, default). The helper uses typed two-branch narrowing — `quorum=0` and `quorum=None` are normalized to literal `None` at the boundary (the docstring documents them as equivalent), and `quorum_timeout` is dropped for the no-quorum branch.

Pool-side validation is limited to what only the pool can see: `quorum > pool capacity` (where capacity is `spawn`, `spawn + lease`, or `lease` depending on mode — extracted as the `_validate_quorum` static helper to avoid triplication across mode arms), `lease == 0` for discovery-only, and `quorum_timeout` supplied with `quorum in (None, 0)`. All remaining quorum/quorum_timeout validation is delegated to `WorkerProxy.__init__` and surfaces at `__aenter__` time when `_make_proxy` constructs the proxy.

### Existing test adjustments

- `test___init___with_zero_lease_and_spawn` and `test___init___with_zero_lease_discovery_only` pass `quorum=0` alongside `lease=0` since the new validation rejects `quorum > lease`.
- The Hypothesis strategy in `test___init___accepts_valid_spawn_and_lease` narrows `lease >= 1` to stay above the default quorum.
- Pool tests that previously asserted construction-time `ValueError` for `quorum < 0`, discovery-only `quorum > lease`, non-positive `quorum_timeout`, and negative-quorum Hypothesis are converted to async `async with WorkerPool(...):` tests asserting `__aenter__`-time `ValueError`, reflecting the pool's delegation to `WorkerProxy`.

## Test cases

| 1 | `TestWorkerProxy` | A discovery service and quorum of -1 | `WorkerProxy` is instantiated | Raise `ValueError` | Negative quorum rejection |
| 2 | `TestWorkerProxy` | A discovery service and quorum of 0 | `WorkerProxy` is instantiated | Create the proxy successfully | Zero quorum acceptance |
| 3 | `TestWorkerProxy` | A discovery service and quorum of 3 | `WorkerProxy` is instantiated | Create the proxy successfully | Positive quorum acceptance |
| 4 | `TestWorkerProxy` | A discovery service with lease=2 and quorum=3 | `WorkerProxy` is instantiated | Raise `ValueError` | Quorum-exceeds-lease rejection |
| 5 | `TestWorkerProxy` | A discovery service with lease=3 and quorum=3 | `WorkerProxy` is instantiated | Create the proxy successfully | Quorum equal to lease acceptance |
| 6 | `TestWorkerProxy` | A discovery service and quorum of 5 with no lease | `WorkerProxy` is instantiated | Create the proxy successfully | Quorum without lease acceptance |
| 7 | `TestWorkerProxy` | A discovery service, quorum=1, quorum_timeout=0 | `WorkerProxy` is instantiated | Raise `ValueError` | Non-positive quorum_timeout rejection |
| 8 | `TestWorkerProxy` | A discovery service, quorum=None, explicit quorum_timeout | `WorkerProxy` is instantiated | Raise `ValueError` | quorum_timeout-without-quorum rejection |
| 9 | `TestWorkerProxy` | A WorkerProxy constructed with quorum=3 and quorum_timeout=15 | `quorum` / `quorum_timeout` properties are read | Return the configured values | Property exposure |
| 10 | `TestWorkerProxy` | A static workers list of length 2 and quorum=5 | `WorkerProxy` is instantiated | Raise `ValueError` matching "compatible worker count" | Static-workers quorum rejection |
| 11 | `TestWorkerProxy` | A static workers list where one worker has an unparseable version, quorum=2 | `WorkerProxy` is instantiated | Raise `ValueError` reporting "1 of 2" after version filtering | Version-filter rejects unparseable |
| 12 | `TestWorkerProxy` | Any lease in [1, 100] and quorum in [0, lease] (Hypothesis) | `WorkerProxy` is instantiated | Create the proxy successfully | Property-based quorum acceptance |
| 13 | `TestWorkerProxy` | A started proxy with quorum=0 and no workers | `dispatch` is called | Proceed without waiting for workers | Zero-quorum dispatch bypass |
| 14 | `TestWorkerProxy` | A started proxy with quorum=2 and a gated discovery stream | `dispatch` is called and workers are added one at a time | Block until 2 workers are discovered | Quorum-gated dispatch blocking |
| 15 | `TestWorkerProxy` | A non-lazy proxy with quorum=2 and unsatisfiable workers | The proxy is entered as `async with` | Raise `asyncio.TimeoutError` at context entry | Non-lazy quorum timeout |
| 16 | `TestWorkerProxy` | A WorkerProxy never started | `stop()` is awaited directly | Raise `RuntimeError` matching "Proxy not started" | Stop-without-start guard |
| 17 | `TestWorkerProxy` | A non-lazy proxy with unsatisfiable quorum and a tiny quorum_timeout | The proxy is used as `async with` (context entry raises) | `wool.__proxy__.get()` returns `None` after the failure | Token reset on start failure |
| 18 | `TestWorkerProxy` | A lazy proxy with unsatisfiable quorum | Two `dispatch()` calls are awaited in sequence | Both raise the same cached `asyncio.TimeoutError` instance | Cached start-failure short-circuit |
| 19 | `TestWorkerProxy` | A non-lazy proxy with quorum=2 pickled and restored | The restored proxy starts and discovers 3 workers | Quorum survives the round-trip | Cloudpickle serialization |
| 20 | `TestWorkerProxy` | A legacy 5-element reduce tuple (no quorum / quorum_timeout) | `_restore_proxy` reconstructs the proxy | Defaults `quorum=1`, `quorum_timeout=60` preserve pre-quorum implicit-wait | Reduce-tuple back-compat |
| 21 | `TestWorkerPool` | A spawn of 4 and quorum of 2 | `WorkerPool` is instantiated | Create the pool successfully | Pool quorum acceptance |
| 22 | `TestWorkerPool` | A spawn of 4 and quorum of 0 | `WorkerPool` is instantiated | Create the pool successfully | Pool zero quorum acceptance |
| 23 | `TestWorkerPool` | A pool with spawn=2, quorum=None, and explicit quorum_timeout | `WorkerPool` is instantiated | Raise `ValueError` | Pool quorum_timeout-without-quorum rejection |
| 24 | `TestWorkerPool` | A spawn of 4 and quorum of -1 | `async with WorkerPool(...)` is entered | Raise `ValueError` from the underlying `WorkerProxy` | Pool negative quorum rejection (delegated) |
| 25 | `TestWorkerPool` | A spawn of 2 and quorum of 5 (no lease) | `WorkerPool` is instantiated | Raise `ValueError` matching "cannot exceed pool capacity" | Pool quorum-exceeds-capacity rejection |
| 26 | `TestWorkerPool` | A pool with spawn=4, lease=3, quorum=7, and discovery | `WorkerPool` is instantiated | Create the pool successfully | Pool quorum equal to capacity acceptance |
| 27 | `TestWorkerPool` | Spawn=4 and quorum=3 (no discovery, no lease) | `WorkerPool` is instantiated | Create the pool successfully | Quorum satisfied by spawn alone |
| 28 | `TestWorkerPool` | A pool with discovery, lease=2, quorum=3 | `async with WorkerPool(...)` is entered | Raise `ValueError` matching "cannot exceed lease" | Discovery-only quorum-exceeds-lease (delegated) |
| 29 | `TestWorkerPool` | A spawn of 2 with no discovery and quorum=10 | `WorkerPool` is instantiated | Raise `ValueError` matching "cannot exceed pool capacity" | Ephemeral rejects quorum > spawn+lease |
| 30 | `TestWorkerPool` | Spawn=2 and quorum=10 with no lease (unbounded) | `WorkerPool` is instantiated | Create the pool successfully (reachability deferred to runtime) | Unbounded lease accepts quorum > spawn |
| 31 | `TestWorkerPool` | A pool with spawn=2, discovery, lease=2, quorum=10 | `WorkerPool` is instantiated | Raise `ValueError` matching "cannot exceed pool capacity" | Hybrid rejects quorum > spawn+lease |
| 32 | `TestWorkerPool` | A pool with spawn=2, quorum=1, quorum_timeout=0 | `async with WorkerPool(...)` is entered | Raise `ValueError` from the underlying `WorkerProxy` | Non-positive quorum_timeout (delegated) |
| 33 | `TestWorkerPool` | A default-mode pool with quorum exceeding `os.cpu_count()` | `WorkerPool` is instantiated | Raise `ValueError` matching "cannot exceed pool capacity" | Default mode quorum-exceeds-capacity |
| 34 | `TestWorkerPool` | A pool with spawn=2 and quorum=2 entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=2` to `WorkerProxy` | Quorum forwarding (explicit) |
| 35 | `TestWorkerPool` | A pool with spawn=2 and no explicit quorum entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=1` to `WorkerProxy` | Quorum forwarding (default) |
| 36 | `TestWorkerPool` | A pool with spawn=2 and explicit `quorum=None` entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=None` to `WorkerProxy` without `quorum_timeout` | _make_proxy no-quorum branch |
| 37 | `TestWorkerPool` | A pool with spawn=2 and `quorum=0` entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=None` (normalized) to `WorkerProxy` without `quorum_timeout` | _make_proxy zero-quorum normalization |
| 38 | `TestWorkerPool` | A pool with discovery and quorum=3 entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=3` to `WorkerProxy` | Durable mode quorum forwarding |
| 39 | `TestWorkerPool` | A pool with spawn=2, discovery, and quorum=3 entered as context manager | `WorkerProxy` is constructed internally | Pass `quorum=3` to `WorkerProxy` | Hybrid mode quorum forwarding |
| 40 | `TestWorkerPool` | Any negative quorum value (Hypothesis) | `async with WorkerPool(...)` is entered | Raise `ValueError` from the underlying `WorkerProxy` | Property-based negative rejection (delegated) |
| 41 | `TestWorkerPool` | Any spawn in [1, 20], lease in [0, 20], quorum in [0, spawn+lease] (Hypothesis) | `WorkerPool` is instantiated | Create the pool successfully | Property-based capacity acceptance |
| 42 | Integration | Pairwise scenarios across discovery × loadbalancer × proxy × quorum dimensions | Pool is run end-to-end | Scenario completes per expected outcome | Quorum dimension in integration scenario matrix |

### DIFF
--- a/wool/src/wool/runtime/worker/pool.py
+++ b/wool/src/wool/runtime/worker/pool.py
@@ -18,12 +18,19 @@ from typing_extensions import deprecated
 from wool.runtime.context import install_task_factory
 from wool.runtime.discovery.base import DiscoveryLike
 from wool.runtime.discovery.base import DiscoveryPublisherLike
+from wool.runtime.discovery.base import DiscoverySubscriberLike
 from wool.runtime.discovery.local import LocalDiscovery
 from wool.runtime.typing import Factory
+from wool.runtime.typing import Undefined
+from wool.runtime.typing import UndefinedType
 from wool.runtime.worker.auth import WorkerCredentials
 from wool.runtime.worker.base import WorkerFactory
 from wool.runtime.worker.base import WorkerLike
 from wool.runtime.worker.local import LocalWorker
+from wool.runtime.worker.proxy import DEFAULT_LAZY
+from wool.runtime.worker.proxy import DEFAULT_QUORUM
+from wool.runtime.worker.proxy import DEFAULT_QUORUM_TIMEOUT
+from wool.runtime.worker.proxy import IneffectiveQuorumTimeoutWarning
 from wool.runtime.worker.proxy import LoadBalancerLike
 from wool.runtime.worker.proxy import RoundRobinLoadBalancer
 from wool.runtime.worker.proxy import WorkerProxy
@@ -138,6 +145,15 @@ class WorkerPool:
         async with WorkerPool(discovery=custom_discovery):
             result = await task()
 
+    **Quorum gate:**
+
+    .. code-block:: python
+
+        # Spawn 4 workers, block on context entry until all 4 are
+        # admitted, and time out after 30s if not.
+        async with WorkerPool(spawn=4, quorum=4, quorum_timeout=30, lazy=False):
+            result = await task()
+
     :param tags:
         Capability tags for spawned workers.
     :param spawn:
@@ -151,14 +167,42 @@ class WorkerPool:
         ``lease`` for external pools. Defaults to ``None`` (unbounded).
     :param worker:
         Worker factory callable. Defaults to :class:`LocalWorker`.
-    :param loadbalancer:
-        Load balancer instance, factory, or context manager.
     :param discovery:
         Discovery service instance, factory, or context manager.
+    :param loadbalancer:
+        Load balancer instance, factory, or context manager.
     :param credentials:
         Optional channel credentials for TLS/mTLS connections to workers.
+    :param quorum:
+        Minimum number of workers that must be discovered before the proxy
+        considers itself ready. Defaults to ``1`` — block until at least
+        one worker is admitted, preserving the pre-quorum implicit-wait
+        behaviour. Pass a larger integer to require more workers, or
+        ``None``/``0`` to disable the gate entirely (``dispatch`` may
+        then raise immediately if no workers have been discovered yet).
+        When ``lazy=True`` (default), the quorum wait is deferred to the
+        first dispatch; with ``lazy=False`` it blocks at context entry.
+    :param quorum_timeout:
+        Seconds to wait for ``quorum`` workers to be discovered before
+        raising :class:`asyncio.TimeoutError`. Only meaningful when
+        ``quorum`` is a positive integer; supplying it with
+        ``quorum=None`` or ``quorum=0`` records the value but never
+        consults it, accompanied by an
+        :class:`~wool.runtime.worker.proxy.IneffectiveQuorumTimeoutWarning`.
+        Defaults to ``60``; pass ``None`` to wait indefinitely. On
+        timeout the pool instance becomes unusable (single-use
+        semantics); construct a new pool to retry.
+    :param lazy:
+        When ``True`` (default), defer discovery setup and the quorum
+        wait to the first :meth:`WorkerProxy.dispatch`.  When ``False``,
+        eagerly enter the underlying proxy at ``__aenter__`` time and
+        run the quorum wait there.
     :raises ValueError:
         If configuration is invalid or CPU count unavailable.
+    :raises asyncio.TimeoutError:
+        If the quorum wait does not complete within ``quorum_timeout``
+        — raised by the underlying :class:`WorkerProxy` at context
+        entry (``lazy=False``) or first dispatch (``lazy=True``).
 
     .. caution::
 
@@ -182,7 +226,9 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None = DEFAULT_QUORUM_TIMEOUT,
+        lazy: bool = DEFAULT_LAZY,
     ):
         """
         Create an ephemeral pool of workers, spawning the specified
@@ -200,10 +246,12 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None = DEFAULT_QUORUM_TIMEOUT,
+        lazy: bool = DEFAULT_LAZY,
     ):
         """
-        Connect to an existing pool of workers discovered by the
+        Connect to an external pool of workers discovered by the
         specified discovery protocol.
         """
         ...
@@ -220,7 +268,9 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None = DEFAULT_QUORUM_TIMEOUT,
+        lazy: bool = DEFAULT_LAZY,
     ):
         """
         Create a hybrid pool that spawns local workers and discovers
@@ -241,7 +291,9 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None = DEFAULT_QUORUM_TIMEOUT,
+        lazy: bool = DEFAULT_LAZY,
     ): ...
 
     @overload
@@ -257,7 +309,9 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None = DEFAULT_QUORUM_TIMEOUT,
+        lazy: bool = DEFAULT_LAZY,
     ): ...
 
     def __init__(
@@ -272,7 +326,9 @@ class WorkerPool:
             LoadBalancerLike | Factory[LoadBalancerLike]
         ) = RoundRobinLoadBalancer,
         credentials: WorkerCredentials | None = None,
-        lazy: bool = True,
+        quorum: int | None = DEFAULT_QUORUM,
+        quorum_timeout: float | None | UndefinedType = Undefined,
+        lazy: bool = DEFAULT_LAZY,
     ):
         self._workers = {}
         self._credentials = credentials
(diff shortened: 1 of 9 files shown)


wool/src/wool/runtime/worker/pool.py
```diff
@@ -155,6 +157,22 @@ async def custom_discovery():
...
+        first dispatch; with ``lazy=False`` it blocks at context entry.
+    :param quorum_timeout:
+        Seconds to wait for ``quorum`` workers to be discovered before
+        raising :class:`asyncio.TimeoutError`. Only meaningful when
+        ``quorum`` is set; ignored otherwise. Defaults to ``60``; pass
+        ``None`` to wait indefinitely. On timeout the pool instance
+        becomes unusable (single-use semantics); construct a new pool
+        to retry.
     :param credentials:
         Optional channel credentials for TLS/mTLS connections to workers.
```
conradbzura: The order of arguments here should match the actual signature of `__init__`.

wool/src/wool/runtime/worker/pool.py
```diff
@@ -208,6 +267,25 @@ def __init__(
...
+        loadbalancer: (
+            LoadBalancerLike | Factory[LoadBalancerLike]
+        ) = RoundRobinLoadBalancer,
+        credentials: WorkerCredentials | None = None,
+        quorum: int = 1,
+        quorum_timeout: float | None = 60,
+        lazy: bool = True,
+    ):
+        """
+        Connect to an existing pool of workers with a quorum gate.
```
conradbzura: "an existing pool" -> "an external pool"

wool/src/wool/runtime/worker/pool.py
```diff
@@ -373,6 +548,15 @@ async def create_proxy():
...
                 max_workers = spawn + lease if lease is not None else None
+                if (
+                    max_workers is not None
+                    and quorum is not None
+                    and quorum > max_workers
+                ):
+                    raise ValueError(
+                        f"Quorum ({quorum}) cannot exceed pool capacity "
+                        f"({max_workers}) — the quorum would never be satisfied"
+                    )
```
conradbzura: This check is repeated three times — hoist into a private static helper method.

wool/src/wool/runtime/worker/pool.py
```diff
@@ -468,6 +653,53 @@ def factory(*tags, credentials=None):
...
+            )
+        return WorkerProxy(
+            discovery=discovery,
+            loadbalancer=loadbalancer,
+            credentials=self._credentials,
+            lease=lease,
+            quorum=quorum,
+            quorum_timeout=quorum_timeout,
+            lazy=lazy,
+        )
```
conradbzura, line 599: Would keyword args make this less verbose?

wool/src/wool/runtime/worker/proxy.py
```diff
@@ -133,15 +139,36 @@ def _restore_proxy(
...
+        )
+    else:
+        proxy = WorkerProxy(
+            discovery=discovery,
+            loadbalancer=loadbalancer,
+            lease=lease,
+            quorum=quorum,
+            quorum_timeout=quorum_timeout,
+            lazy=lazy,
+        )
```
conradbzura, line 194: Again, would keyword args make this less verbose?

PR #157, opened 2026-04-06
Implement inter-worker context propagation — Closes #154, Fixes #176

conradbzura:
## Summary

Add `wool.ContextVar` — a propagating context variable that mirrors the standard library's `contextvars.ContextVar` surface (`get`, `set`, `reset` with a `Token`) and ships values across worker boundaries on every dispatch frame. Identity is a `"namespace:name"` key inferred from the caller's top-level package; state lives on a per-task `wool.Context` that propagates bidirectionally on every gRPC `Request` and `Response` frame; reset signals propagate alongside, riding as hex-encoded token ids inlined under each var's wire entry.

The work centers on three deliberate divergences from the standard library: cross-process serialization, since `contextvars.ContextVar` and `contextvars.Context` are C-extensions that block pickle; a strict registry-backed identity model, where key collisions raise instead of producing two distinct vars under the same name; and a single-task-per-`wool.Context` invariant that tracks each chain of execution as a first-class object with a UUID. Bidirectional propagation runs inside the existing dispatch flow with per-entry decode resilience: corrupt vars surface a `ContextDecodeWarning`, the entry is skipped, the routine still runs, and the response still ships. Strict mode (`PYTHONWARNINGS=error::wool.ContextDecodeWarning`) promotes the warning to an exception and aggregates per-var failures into a single `BaseExceptionGroup`.

The branch also fixes a pre-existing bug on the async-generator dispatch path. The streaming worker loop ran the generator outside the task's `RuntimeContext`, dropping `dispatch_timeout` and any nested-dispatch state. The coroutine path already handled this correctly via `Task._run`.

Closes #154, Fixes #176.

## Design rationale

**Global identity via namespaced key.** A `wool.ContextVar` is uniquely identified across the cluster by its `"namespace:name"` key. Standard library `contextvars.ContextVar` instances are identified by Python's module-level singleton convention: declare once at module scope, refer to the same instance throughout. Wool formalizes that convention as a registry invariant — two `ContextVar(...)` calls with the same key raise `ContextVarCollision`, even if the second declaration happens to supply a matching default. The collision check preserves the module-scoped pattern callers already use and makes cross-process identity well-defined.

[...]

PR #193, opened 2026-05-02
Restrict pickling of Wool types to Wool's internal pickler — Closes #192

### LINKED ISSUE #192: Restrict pickling of Wool types to Wool's internal pickler
## Description

Wool has a growing set of custom picklable types whose `__reduce__` implementations only have meaningful semantics when paired with Wool's own dispatch machinery on the receiving side — manifest table, task-keyed store, matching runtime, and so on. Today, because these types implement the standard `__reduce__` protocol, vanilla `pickle.dumps` and `cloudpickle.dumps` will happily serialize them, producing payloads that deserialize into nonsensical state outside a Wool worker.

This refactor replaces the public protocol with a Wool-private one and routes Wool's own serialization through a dedicated pickler:

1. Introduce `WoolPickler(cloudpickle.Pickler)` (and `WoolUnpickler` for symmetry if needed). `WoolPickler.reducer_override(obj)` returns `obj.__wool_reduce__()` when the method exists and `NotImplemented` otherwise.
2. Establish the convention: a Wool type that needs custom pickling defines `__wool_reduce__` (returning the same `(callable, args)` tuple shape as `__reduce__`) and `__reduce_ex__` that raises `TypeError` with a message naming Wool's dispatcher as the only supported serialization path.
3. Identify every Wool type on `main` that currently defines `__reduce__` and migrate it to the new pattern: rename `__reduce__` → `__wool_reduce__`, add the `__reduce_ex__` guard.
4. Route every dispatch-path serialization site that currently calls into cloudpickle through `WoolPickler`.

Because `WoolPickler.reducer_override` duck-types on the presence of `__wool_reduce__`, future picklable types opt in by implementing the method — no per-type changes to the pickler.

## Motivation

Silent failures from accidental cross-boundary pickling are difficult to diagnose: a Wool object placed in a dataclass that someone later `copy.deepcopy`s, shoved into a `multiprocessing.Queue`, or pickled by an application-level cache currently succeeds at serialization time and fails much later, far from the cause, when the deserialized object is used outside a Wool runtime. The current `__reduce__` definitions implicitly advertise these types as freely picklable, which they are not.

Failing loudly at the pickle boundary moves the error to the moment of misuse and makes Wool's serialization contract explicit: these types are Wool-internal wire formats, not general-purpose serializable objects.

## Expected Outcome

- `pickle.dumps(x)` and `cloudpickle.dumps(x)` raise `TypeError` for any `x` whose type is on the migrated list, with a message identifying Wool's dispatcher as the supported path.
- Wool's dispatch pipeline serializes and deserializes these types unchanged from current behavior.
- Adding a new Wool-picklable type requires only implementing `__wool_reduce__` on that type — no edits to `WoolPickler`.

conradbzura:
## Summary

Introduce a Wool-internal pickling protocol — `__wool_reduce__` for serialization through Wool's own pickler and `__reduce_ex__` to reject vanilla `pickle` and `cloudpickle` — and apply it to `WorkerProxy`. The new `wool.runtime.serializer` module hosts the protocol mechanism, the `Serializer` Protocol (moved from `wool.runtime.routine.task`), the `CloudPickleSerializer` default implementation, and the module-level `DEFAULT_SERIALIZER` singleton consumed by every dispatch site. Wool types that hold runtime-bound dispatch state — initially just `WorkerProxy`, later the `wool.Context` / `wool.ContextVar` / `wool.Token` types from #157 — fail loudly when serialized outside the dispatch path. The receiving side continues to use `cloudpickle.loads` because `__wool_reduce__` returns standard `(callable, args)` reduce tuples.

Other picklable Wool types (`RoundRobinLoadBalancer`, `LocalDiscovery`, `afilter`, `ReducibleAsyncIterator`, `PassthroughSerializer`) implement user-facing protocols whose user implementations must be cloudpickle-picklable. Wool's own implementations of those protocols stay freely picklable to match that contract.

Closes #192.

## Design rationale

**Two categories of picklable Wool types, with only one guarded.** Wool's picklable types split into dispatch-state types (`WorkerProxy` today; `wool.Context` / `wool.ContextVar` / `wool.Token` once #157 lands) and protocol implementations (`RoundRobinLoadBalancer`, `LocalDiscovery`, `afilter`, `PassthroughSerializer`, `ReducibleAsyncIterator`). Dispatch-state types hold runtime-bound state — credentials resolved against the active `CredentialContext`, manifest tokens against the task-keyed store — that is meaningless outside a Wool runtime. Protocol implementations carry no runtime-bound state; user implementations of the same protocols must be cloudpickle-picklable, and Wool's own implementations follow that contract by symmetry. Guarding the protocol implementations would force a divergent rule on Wool's implementations versus user-supplied ones with no corresponding benefit.

**`reducer_override` over `copyreg.dispatch_table`.** The `_WoolPickler` subclass overrides `cloudpickle.Pickler.reducer_override(obj)` and looks for `__wool_reduce__` on the object's type; when absent, the override delegates to `super().reducer_override(obj)` so cloudpickle's special-case handling for lambdas, locally-defined classes, and other dynamic types is preserved. A `copyreg`-style dispatch table was rejected because it requires every picklable type to be registered in a single mapping — coupling the pickler to the set of opt-in types and forcing edits to a central file every time a new type is added. Duck-typing on `__wool_reduce__` keeps the opt-in on the type itself.

**`__reduce_ex__` carries the guard, not `__reduce__`.** Cloudpickle's pickler calls `__reduce_ex__` first, falling back to `__reduce__` if the former returns `NotImplemented`; raising in `__reduce_ex__` short-circuits both paths. Because `reducer_override` is consulted before `__reduce_ex__`, the guard is invisible to `_WoolPickler` — the override fires and serialization proceeds via `__wool_reduce__`. The guard fires for vanilla picklers and stays out of Wool's way. `copy.copy` and `copy.deepcopy` route through `__reduce_ex__` as well, so they are rejected through the same path — a runtime-bound proxy has no meaningful copy semantics.

**`__wool_reduce__` returns the standard reduce-tuple shape.** The unpickling side does not need Wool-specific machinery to reconstruct the object; the standard pickle reconstructor runs the callable with the args as it would for any reduce tuple. This keeps the migration trivial — rename the existing `__reduce__` method — and means there is no `WoolUnpickler`. Cloudpickle ships only a `Pickler`, not an `Unpickler`, for the same reason: serialization is the only side that needs custom logic.

**Default `Serializer` becomes `CloudPickleSerializer`, owned in `wool.runtime.serializer`.** The pre-change fallback at `task.py:358` was bare `cloudpickle.dumps`, which after the migration would fail any time a guarded type appeared in a Task payload. `CloudPickleSerializer` is a `Serializer` Protocol implementation that uses `_WoolPickler` underneath; the module-level `DEFAULT_SERIALIZER` singleton is the canonical instance. Every dispatch site — `task.py`'s default fallback, `worker/connection.py`, `worker/service.py`, `worker/process.py` — imports the same singleton. End-user-facing semantics remain "your callables and args must be cloudpickle-picklable"; the `_WoolPickler` aspect is an implementation detail.

**`Serializer` Protocol requires hashability.** Wool caches the pickled form of each serializer instance via `functools.lru_cache`; an implementation that overrides `__eq__` without supplying a compatible `__hash__` (which Python silently sets to `None`) would trip the cache with a `TypeError` on first dispatch. The Protocol declares `__hash__(self) -> int` so static type checkers catch this. Runtime `isinstance` cannot enforce it (Python's `hasattr` returns True for `__hash__ = None`), so the docstring states the contract explicitly.

**`_WoolPickler` is module-private.** Its only consumer is `CloudPickleSerializer.dumps`. The leading underscore signals that callers should not subclass it directly; the surface they consume is the `Serializer` Protocol, the `CloudPickleSerializer` class, or the `DEFAULT_SERIALIZER` singleton.

### `wool.runtime.serializer` (new module)

Houses the serialization surface lifted out of `wool.runtime.routine.task`:

- `Serializer` Protocol — moved verbatim from `task.py` and extended with `__hash__(self) -> int`. Re-exported via `wool.__all__` under a new "Serialization" group.
- `_WoolPickler` — `cloudpickle.Pickler` subclass whose `reducer_override` honors `__wool_reduce__` when present and delegates to `super()` otherwise.
- `CloudPickleSerializer` — `Serializer` implementation. `dumps` wraps `_WoolPickler` over a `BytesIO` buffer; `loads` delegates to `cloudpickle.loads`. All instances hash and compare equal so the `_pickle_serializer` LRU cache collapses them to one slot.
- `DEFAULT_SERIALIZER` — module-level `CloudPickleSerializer` singleton. The canonical default for dispatch-path serialization.
- `_PassthroughKey`, `_passthrough_store`, `PassthroughSerializer` — moved verbatim from `task.py`. Behavior unchanged.

### `WorkerProxy` migration (`wool/runtime/worker/proxy.py`)

Rename `WorkerProxy.__reduce__` to `__wool_reduce__`; preserve its existing body — the loadbalancer/discovery context-manager check, the credential exclusion, the proxy-id preservation. Add `WorkerProxy.__reduce_ex__` raising `TypeError` with a message that names Wool's runtime as the supported serialization path. Hoist the previously-nested `_restore_proxy` helper to module level so the reduce tuple references a stable callable rather than a freshly-created closure on every reduction. The hoist matches the convention established by `_proxy_factory` in `worker/process.py`.

### Dispatch-path routing

`wool/runtime/routine/task.py` — drop the local `_DEFAULT_SERIALIZER` declaration; import `DEFAULT_SERIALIZER` from `wool.runtime.serializer`. The `Task.to_protobuf` `dumps` and `proxy_dumps` fallbacks resolve to `DEFAULT_SERIALIZER.dumps` when no explicit serializer is supplied. `_pickle_serializer` (which serializes the `Serializer` instance itself for the wire) routes through `DEFAULT_SERIALIZER.dumps` for consistency.

`wool/runtime/worker/connection.py`, `worker/service.py`, `worker/process.py` — replace direct `cloudpickle.dumps` calls with `DEFAULT_SERIALIZER.dumps`. The `cloudpickle.loads` calls stay unchanged.

### `wool/__init__.py`

Re-export `Serializer` from the new `wool.runtime.serializer` location. Move it under a new "Serialization" group in `__all__` (was grouped under "Routines"); update `tests/test_public.py`'s expected ordering accordingly.

## Test cases

| 1 | `TestCloudPickleSerializer` | A guarded type defining `__wool_reduce__` and a `__reduce_ex__` guard | The serializer dumps and loads the instance | The result equals the original | Guarded-type round-trip |
| 2 | `TestCloudPickleSerializer` | A simple string payload | The serializer dumps and loads it | The result equals the original | Standard payload round-trip |
| 3 | `TestCloudPickleSerializer` | A locally-defined lambda | The serializer dumps and loads it | The restored callable returns the same value | Cloudpickle delegation preserved |
| 4 | `TestCloudPickleSerializer` | A guarded type instance | `pickle.dumps` and `cloudpickle.dumps` are called on it | Both raise `TypeError` from the `__reduce_ex__` guard | Vanilla pickle rejection |
| 5 | `TestCloudPickleSerializer` | Arbitrary text, integer, binary, or list payloads (Hypothesis) | The serializer dumps and loads each | Each round-trips to an equal value | Property-based round-trip |
| 6 | `TestCloudPickleSerializer` | Two distinct serializer instances | Their hashes are computed | The hashes are equal so the LRU cache deduplicates them | Cache compatibility |
| 7 | `TestCloudPickleSerializer` | Two distinct serializer instances | They are compared for equality | They compare equal so the cache treats them as one key | Cache compatibility |
| 8 | `TestCloudPickleSerializer` | A `CloudPickleSerializer` and an arbitrary other object | They are compared for equality | They do not compare equal | Cross-type equality |
| 9 | `TestDefaultSerializer` | The module-level `DEFAULT_SERIALIZER` constant | Its type is inspected | It is a `CloudPickleSerializer` instance | Singleton identity |
| 10 | `TestWorkerProxy` | A `WorkerProxy` instance | `pickle.dumps` and `cloudpickle.dumps` are called on it directly | Both raise `TypeError` naming Wool's runtime as the supported path | Vanilla pickle rejection |
| 11 | `TestTask` | A `Task` whose proxy implements `__wool_reduce__` and the `__reduce_ex__` guard | `to_protobuf` is called with no explicit `Serializer` | `DEFAULT_SERIALIZER` serializes the proxy successfully via `_WoolPickler` | Default Serializer respects guarded types |

Existing `WorkerProxy` pickle round-trip tests in `tests/runtime/worker/test_proxy.py` (~14 sites) were rewired from `cloudpickle.dumps(proxy)` to `DEFAULT_SERIALIZER.dumps(proxy)` so they continue to verify round-trip behavior through the migrated pickler. The previously-existing `TestWoolPickler` tests were removed when `WoolPickler` became private; their behaviors are covered by `TestCloudPickleSerializer` end-to-end.

### DIFF
--- a/wool/src/wool/runtime/serializer.py
+++ b/wool/src/wool/runtime/serializer.py
@@ -1,0 +1,178 @@
+from __future__ import annotations
+
+import io
+import weakref
+from typing import Any
+from typing import Protocol
+from typing import runtime_checkable
+from uuid import UUID
+from uuid import uuid4
+
+import cloudpickle
+
+
+# public
+@runtime_checkable
+class Serializer(Protocol):
+    """Protocol for pluggable serialization of dispatch payloads.
+
+    Implementations must be hashable.  Wool caches the pickled form of each
+    serializer instance via :func:`functools.lru_cache`.  A class that
+    overrides ``__eq__`` without supplying a compatible ``__hash__``
+    (which Python silently sets to ``None``) fails ``isinstance(obj,
+    Serializer)`` because the runtime-checkable protocol consults
+    ``hasattr``, and ``hasattr(obj, "__hash__")`` returns ``False`` when
+    ``__hash__`` is ``None``.  An instance that bypasses both static
+    type checking and the ``isinstance`` gate would trip the cache with
+    a :exc:`TypeError` on first dispatch.
+    """
+
+    def __hash__(self) -> int: ...
+
+    def dumps(self, obj: Any) -> bytes:
+        """Serialize *obj* to bytes for transport across a worker boundary.
+
+        :param obj:
+            The object to serialize.  Implementations decide which payloads
+            are supported.
+        :returns:
+            The serialized representation of *obj*.
+        """
+        ...
+
+    def loads(self, data: bytes) -> Any:
+        """Deserialize *data* produced by a matching :meth:`dumps` call.
+
+        :param data:
+            Bytes previously produced by :meth:`dumps`.
+        :returns:
+            The reconstructed object.
+        """
+        ...
+
+
+class _WoolPickler(cloudpickle.Pickler):
+    """Cloudpickle-based pickler that honors the ``__wool_reduce__`` protocol.
+
+    For objects whose type defines ``__wool_reduce__``, the method's return
+    value (a standard ``(callable, args)`` reduce tuple) is used in place of
+    the standard reduction protocol.  All other objects are handled exactly
+    as :class:`cloudpickle.Pickler` would handle them.
+
+    The override fires before ``__reduce_ex__``, so a type that pairs
+    ``__wool_reduce__`` with a ``__reduce_ex__`` guard (raising
+    :exc:`TypeError`) can be guarded against vanilla pickling while still
+    serializing correctly through this pickler.
+    """
+
+    def reducer_override(self, obj: Any) -> Any:
+        if hasattr(type(obj), "__wool_reduce__"):
+            return obj.__wool_reduce__()
+        return super().reducer_override(obj)
+
+
+class CloudpickleSerializer:
+    """Default :class:`Serializer` implementation.
+
+    Used by :class:`~wool.runtime.routine.task.Task.to_protobuf` when the
+    caller does not provide an explicit serializer.  Semantically the
+    contract is cloudpickle's: callables and arguments must be
+    cloudpickle-picklable.  The fact that ``dumps`` routes through Wool's
+    internal pickler — and therefore respects the ``__wool_reduce__``
+    protocol on guarded Wool types — is an implementation detail.
+    """
+
+    def __hash__(self) -> int:
+        """Return a constant hash so all instances share an LRU-cache slot."""
+        return hash(CloudpickleSerializer)
+
+    def __eq__(self, other: object) -> bool:
+        """Return True for any other CloudpickleSerializer; instances are interchangeable."""
+        return isinstance(other, CloudpickleSerializer)
+
+    def dumps(self, obj: Any) -> bytes:
+        """Serialize *obj* via Wool's internal pickler.
+
+        Honors the ``__wool_reduce__`` protocol on guarded Wool types and
+        otherwise reduces via cloudpickle.
+        """
+        buffer = io.BytesIO()
+        _WoolPickler(buffer).dump(obj)
+        return buffer.getvalue()
+
+    def loads(self, data: bytes) -> Any:
+        """Deserialize *data* via :func:`cloudpickle.loads`."""
+        return cloudpickle.loads(data)
+
+
+class _PassthroughKey:
+    """Weak-referenceable key for the passthrough store."""
+
+    __slots__ = ("__weakref__", "token")
+
+    def __init__(self, token: UUID | None = None) -> None:
+        self.token = token if token is not None else uuid4()
+
+    def __hash__(self) -> int:
+        return hash(self.token)
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, _PassthroughKey):
+            return super().__eq__(other)
+        return self.token == other.token
+
+
+# Weak keys let the owning PassthroughSerializer's `_keys` list act as the
+# lifetime anchor: when the serializer goes out of scope, its keys are
+# garbage-collected and the corresponding entries here are removed
+# automatically without any explicit cleanup.
+_passthrough_store: weakref.WeakKeyDictionary[_PassthroughKey, Any] = (
+    weakref.WeakKeyDictionary()
+)
+
+
+class PassthroughSerializer:
+    """In-process serializer that avoids pickling entirely.
+
+    Each instance acts as a scope guard for one dispatch.
+    ``dumps`` creates a weakly-referenceable key, stores the object
+    in a module-level :class:`~weakref.WeakKeyDictionary`, and
+    retains a strong reference to the key on ``self``.  When the
+    serializer goes out of scope the keys are garbage-collected and
+    the weak-dict entries are removed automatically.
+
+    ``loads`` is static — it reconstructs the key from the bytes
+    token in the protobuf message and pops the entry from the store.
+
+    All instances hash and compare equal so the LRU cache that pickles
+    serializer instances for transport hits on every call.
+    """
+
+    def __init__(self) -> None:
+        """Initialize an empty list of strong key references for this dispatch scope."""
+        self._keys: list[_PassthroughKey] = []
+
+    def __hash__(self) -> int:
+        """Return a constant hash so all instances share an LRU-cache slot."""
+        return hash(PassthroughSerializer)
+
+    def __eq__(self, other: object) -> bool:
+        """Return True for any other PassthroughSerializer; instances are interchangeable."""
+        return isinstance(other, PassthroughSerializer)
+
+    def __reduce__(self) -> tuple[type[PassthroughSerializer], tuple[()]]:
+        """Reduce to a fresh instance; the per-scope key list does not survive transport."""
+        return (PassthroughSerializer, ())
+
+    def dumps(self, obj: Any) -> bytes:
+        """Stash *obj* in the module store and return its weak-key token."""
+        key = _PassthroughKey()
+        self._keys.append(key)
+        _passthrough_store[key] = obj
+        return key.token.bytes
+
+    @staticmethod
+    def loads(data: bytes) -> Any:
+        """Pop and return the object stashed under the token in *data*."""
+        key = _PassthroughKey(UUID(bytes=data))
+        return _passthrough_store.pop(key)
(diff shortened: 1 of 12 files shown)


wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,171 @@
...
+    def dumps(self, obj: Any) -> bytes:
+        buffer = io.BytesIO()
+        _WoolPickler(buffer).dump(obj)
+        return buffer.getvalue()
+
+    def loads(self, data: bytes) -> Any:
+        return cloudpickle.loads(data)
+
+
+DEFAULT_SERIALIZER: Final[CloudPickleSerializer] = CloudPickleSerializer()
```
conradbzura: We should move this to `wool/src/wool/__init__.py` and call is `__serializer__`. The docstring at `.sandbox/issue-154/wool/src/wool/runtime/routine/task.py:277` will need to be updated as a result. Audit all docstrings and comments touched by this PR for staleness after this change as well.

wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,167 @@
...
+from uuid import UUID
+from uuid import uuid4
+
+import cloudpickle
+
+
+# public
+@runtime_checkable
+class Serializer(Protocol):
+    """Protocol for pluggable serialization of Task payload fields.
```
conradbzura: It won't just be tasks for long. Make this more generic — "serialization of dispatch frames" or something like that.

wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Implementations MUST be hashable.  Wool caches the pickled form of each
```
conradbzura: "MUST" -> "must".

wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,167 @@
...
+    internal pickler — and therefore respects the ``__wool_reduce__``
+    protocol on guarded Wool types — is an implementation detail.
+    """
+
+    def __hash__(self) -> int:
+        return hash(CloudPickleSerializer)
+
+    def __eq__(self, other: object) -> bool:
+        # All instances collapse to one cache slot in
+        # _pickle_serializer's lru_cache (mirrors PassthroughSerializer).
```
conradbzura: Drop the comment.

wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,167 @@
...
+        return isinstance(other, CloudPickleSerializer)
+
+    def dumps(self, obj: Any) -> bytes:
+        buffer = io.BytesIO()
+        _WoolPickler(buffer).dump(obj)
+        return buffer.getvalue()
+
+    def loads(self, data: bytes) -> Any:
+        return cloudpickle.loads(data)
```
conradbzura, line 105: These methods could use docstrings.

wool/src/wool/runtime/serializer.py
```diff
@@ -0,0 +1,167 @@
...
+    def dumps(self, obj: Any) -> bytes:
+        key = _PassthroughKey()
+        self._keys.append(key)
+        _passthrough_store[key] = obj
+        return key.token.bytes
+
+    @staticmethod
+    def loads(data: bytes) -> Any:
+        key = _PassthroughKey(UUID(bytes=data))
+        return _passthrough_store.pop(key)
```
conradbzura, line 178: These methods could use docstrings.

PR #205, opened 2026-05-11
Unify worker-side coroutine and async-generator dispatch paths via a single driver — Closes #187

### LINKED ISSUE #180: Consume Task._stream() in WorkerService streaming dispatch path
## Description

WorkerService._stream_from_worker.worker_dispatch in wool/src/wool/runtime/worker/service.py reimplements generator-iteration logic that Task._stream already encapsulates: RuntimeContext/Task context-manager entry, do_dispatch framing, proxy handling, and aclose teardown. The two implementations have drifted — PR #157 re-added the missing `with work_task.context:` wrap only because the inline loop had fallen out of sync with Task._stream. Consume Task._stream() directly so the two paths cannot diverge again.

The obstacle is that worker_dispatch runs on a dedicated worker event loop and must accept caller-driven next/send/throw commands over an asyncio.Queue, whereas Task._stream currently exposes a plain `async for` iterator with no hook for external driving. The refactor therefore requires either:

1. Adding a queue/command-driven iteration API to Task that produces the same per-frame wrap as the plain iterator, or
2. Switching the cross-loop coordination primitive so the caller drives Task._stream directly rather than through a side queue.

## Motivation

The current duplication is the proximate cause of at least one regression already fixed in #157. Each future change to the task execution protocol — RuntimeContext, do_dispatch, proxy plumbing, context snapshotting — has to be applied in two places or the inline loop will silently drift again. Single-point-of-truth in Task eliminates the class of bug.

## Expected Outcome

- WorkerService streaming dispatch consumes Task._stream() (directly or via a new queue-driven iteration hook added to Task).
- The inline generator-driver loop in worker_dispatch is removed.
- Streaming and coroutine dispatch paths share the same RuntimeContext/Task/do_dispatch wrap, enforced structurally rather than by convention.
- No behavior change observable to callers.

### LINKED ISSUE #181: Await worker task unwind on streaming dispatch cancellation
## Description

WorkerService._stream_from_worker's finally block schedules _STREAM_END on the worker loop's request_queue but does not await the worker task's unwind. When the dispatch handler's streaming generator is cancelled (client disconnect mid-stream, deadline exceeded, etc.), control returns to the handler's fallback at wool/src/wool/runtime/worker/service.py:356 while the worker task may still be mid-unwind on its own loop. The fallback reads handler_ctx._data on the gRPC handler loop; the worker may still be mutating that dict.

Under CPython, concurrent iteration and mutation of a dict can raise `RuntimeError: dictionary changed size during iteration` from the iterator side, which would mask the originating CancelledError and surface a spurious handler-level failure to the caller. A torn snapshot (partial mutations) is also possible even when the RuntimeError does not fire.

PR #157 closed the equivalent race on the coroutine dispatch path by awaiting the worker's completion future in _run_on_worker's CancelledError branch. The streaming path has no analogous plumbing: the worker task is created inside _stream_from_worker._start_worker, is not saved to a handle accessible outside that closure, and lives on a different event loop from the generator.

## Expected Behavior

When a streaming dispatch is cancelled at the handler, the worker-loop task has finished unwinding before the dispatch handler's fallback reads handler_ctx._data. The caller's error Response either ships a clean best-effort snapshot or omits vars entirely — never a torn dict or a RuntimeError shadowing the originating exception.

## Root Cause

wool/src/wool/runtime/worker/service.py:682-683:

```python
finally:
    worker_loop.call_soon_threadsafe(request_queue.put_nowait, _STREAM_END)
```

The finally only schedules the stop signal and returns. No handle to the worker asyncio.Task is kept in scope, and there is no bridge to a concurrent.futures.Future the handler loop could await on.

Fix shape: save the worker task in a nonlocal inside _stream_from_worker, attach a done callback that sets a concurrent.futures.Future when it fires, and in the finally block `await asyncio.wrap_future(done_future)` after scheduling _STREAM_END. That matches the coroutine path's approach and requires no cross-loop locking. Alternatively, fold the worker task handle into the refactor tracked in #180 (consume Task._stream directly) where the duplicate iteration loop is already being eliminated — the two issues share most of the underlying plumbing.

### LINKED ISSUE #187: Unify worker-side coroutine and async-generator dispatch paths via a single _drive_routine driver
## Description

`WorkerService._run_on_worker` (coroutine path) and `WorkerService._stream_from_worker` (async-generator path) in `wool/src/wool/runtime/worker/service.py` are two parallel implementations of "wool task on the worker loop" that share roughly 70% of their logic but no code.

Reframe both shapes as iteration of a unified routine driver. A coroutine is a one-frame iterator; an async generator is an N-frame iterator. The driver:

1. Acquires the worker loop and sets up the cross-loop result queue.
2. Calls a single `_drive_routine(work_task, work_ctx, command_stream, *, serializer)` coroutine on the worker loop. For coroutines the command stream is a synthetic single `("next", None, None)`; for generators it is the real request queue fed from the main loop.
3. Per command, runs the routine step inside the wool-task scope (`with task.context, with task, with do_dispatch(False):`) — the scope is established **once**, around the whole loop, matching today's streaming path's #176 fix.
4. Builds a `_WorkerOutcome` per step via `_outcome_from`, puts it on the result queue.

End state: `_run_on_worker` and `_stream_from_worker` collapse into one function. The wool-task scaffolding (`with task.context, with task, with do_dispatch(False):`) lives in one place on the worker side — today it exists in `Task._run` (caller-side) and is reimplemented inside the worker-side `worker_dispatch` closure.

## Motivation

The duplication has been a recurring bug surface:

- A regression where `Task.__enter__` was placed inside the streaming path's per-iteration loop, because the streaming path reimplemented `Task._run`'s setup and the two diverged.
- An embedded `wool.ContextVar` reconstitution bug whose fix had to land at four mirrored sites (initial-frame decode, mid-stream decode, coroutine snapshot encode, streaming snapshot encode).
- Worker-side forward propagation of `consumed_tokens` losing state on second-hop dispatches.
- The decode/encode failure semantics (primary/ancillary split with `ContextDecodeWarning` and `__notes__`) needed to be applied at four mirrored sites.
- The snapshot-encode + warn + `add_note` shape duplicated across both paths.

Two earlier rounds of consolidation already landed in PR #157, introducing the helpers that this issue's `_drive_routine` builds on:

- `_decode_context` — raises on decode failure, callers catch and warn.
- `_snapshot_with_fallback(ctx, *, serializer, primary_exc) -> protocol.Context` — encode + warn + add_note + fallback.
- `_outcome_from(*, value, routine_exc, ctx, serializer) -> _WorkerOutcome` — combines snapshot + result-vs-exception fork in one place.
- The streaming worker loop now puts `_WorkerOutcome` directly on the result queue (no more `(tag, payload, wire_ctx)` intermediary).

This issue is the structural step that prevents the next instance of this bug class from landing as parallel fixes at multiple mirrored sites.

## Expected Outcome

- `_run_on_worker` and `_stream_from_worker` are gone, replaced by a single dispatch entry that delegates to `_drive_routine`.
- The wool-task scope is established once per dispatch on the worker side, regardless of routine shape.
- Streaming `worker_dispatch` (~175 lines today) shrinks dramatically.
- Estimated diff: ~−150 lines net on `service.py`.
- Existing tests pass without modification:
  - `tests/runtime/worker/test_service.py::test_dispatch_with_unpicklable_worker_mutation`
  - `tests/runtime/worker/test_service.py::test_dispatch_streaming_with_unpicklable_worker_mutation`
  - `tests/runtime/worker/test_service.py::test_dispatch_with_routine_raise_and_unpicklable_mutation`
  - `tests/runtime/worker/test_service.py::test_dispatch_streaming_with_routine_raise_and_unpicklable_mutation`
  - `tests/runtime/worker/test_service.py::test_dispatch_streaming_with_mid_stream_corrupt_context`
  - `tests/runtime/worker/test_service.py::test_dispatch_streaming_when_update_raises`
  - All integration tests under `tests/integration/test_context_var_propagation.py`.
- The single-task-per-`wool.Context` invariant remains exclusive to the worker-loop task — the dispatch handler must not hold the Context's running-flag while the routine executes.

**Out of scope but worth evaluating during implementation:** whether `Task._stream` and `Task._run` in `routine/task.py` (used by `Task.dispatch()` on the caller side) should also collapse into a unified iterator. The caller side has different semantics — no command queue, no cross-loop bridging — so this may stay distinct.

**Pre-req.** PR #157 merged.

conradbzura:
## Summary

Unify the worker-side dispatch path by replacing the forked `_run_on_worker` (coroutine) and `_stream_from_worker` (async-generator) implementations with a single `DispatchSession` driver that handles both routine shapes as iteration over a uniform request/response loop. The session is an async context manager and async iterator: `__aenter__` runs the parse phase, `__aiter__` schedules the worker driver lazily on first iteration, and `__aexit__` drains. Extract the routine-scope establishment from `Task._run` / `Task._stream` into a single `routine_scope` helper so caller-side and worker-side execution share one source of truth for proxy acquisition, `wool.__proxy__` binding, runtime-context entry, `do_dispatch` toggle, and async-generator `aclose` teardown.

Consolidate the caller side around the same axis. `WorkerConnection.dispatch` is now a single `AsyncExitStack`-owned streaming generator: the passthrough serializer ref, the channel ref, the concurrency permit, and the gRPC call's cancel hook all sit on one stack with acquire-then-register-release running back-to-back so cancellation cannot interleave between them. The split-stack handoff across the priming yield is gone.

Refine the wire-error classification and cancellation observable. Version mismatch demotes from an application-level `Nack` to a gRPC `FAILED_PRECONDITION` abort — version handshake is part of transport, not parse. Parse-phase rejections keep the `Nack`-with-exception channel so the caller deserializes the original failure class instead of an opaque `RpcError`. Caller-side `_read_next` lets routine-self-raised `asyncio.CancelledError` propagate raw to mirror stdlib's `await task` parity, and degrades non-`CancelledError` `BaseException` payloads to `UnexpectedResponse` instead of `RpcError` so the load balancer does not evict a worker for a routine-level fault. Operator-initiated worker shutdown ships the worker-side `CancelledError` raw on the same routine-failure path as caller cancel — caller-side cancel and operator preempt are now wire-indistinguishable, mirroring stdlib's `await task` where `task.cancel()` from any source produces the same observable.

Closes #187.

## Dispatch flow

```mermaid
sequenceDiagram

    participant Caller
    participant Connection

    participant Service
    participant Session
    participant Worker


    %% -- 1. Handshake --
    rect rgb(0, 0, 0, 0)
        Note over Caller, Worker: Handshake

        Caller ->> Connection: invoke routine (await / async for)
        activate Caller
        Connection ->> Connection: serialize Task to protobuf
        Connection ->> Service: open bidi stream, write Request(task)

        Service ->> Service: VersionInterceptor — parse TaskEnvelope
        alt Incompatible or unparseable version
            Service -->> Connection: abort FAILED_PRECONDITION
            Connection -->> Caller: raise RpcError
        else Compatible
            Service ->> Session: forward Request stream
            activate Session
            Session ->> Session: parse Task, negotiate serializer, decode wool.Context
            alt Parse failure
                Session -->> Service: Rejected(original)
                Service -->> Connection: Nack(exception=original)
                Connection -->> Caller: raise original exception class
            else Parse success
                Session -->> Service: Ack(version)
                Service -->> Connection: Ack(version)
                Connection -->> Caller: ack received
            end
        end
    end

    %% -- 2. Unified iteration --
    rect rgb(0, 0, 0, 0)
        Note over Caller, Worker: Unified driver iteration

        Session ->> Worker: schedule _run on worker loop (lazy, first __aiter__)
        activate Worker
        Worker ->> Worker: enter routine scope — proxy, context, do_dispatch=False

        loop One step per Request (coroutine synthesizes a single "next")
            Caller ->> Connection: next / asend / athrow
            Connection ->> Service: Request(verb, payload, caller context)
            Service ->> Session: forward Request
            Session ->> Worker: enqueue request (cross-loop)
            Worker ->> Worker: _step — decode caller context, advance routine
            alt Step yields a value
                Worker -->> Session: _Response(result, post-step context)
                Session -->> Service: Response(result, context)
                Service -->> Connection: Response(result, context)
                Connection ->> Connection: apply context, deserialize result
                Connection -->> Caller: yield result
            else Generator exhausted
                Worker -->> Session: StopAsyncIteration
                Session -->> Service: Response(exception=StopAsyncIteration)
                Service -->> Connection: Response(exception=StopAsyncIteration)
                Connection -->> Caller: raise StopAsyncIteration
            else Routine raises
                Worker -->> Session: routine exception
                Session -->> Service: Response(exception)
                Service -->> Connection: Response(exception)
                Connection -->> Caller: re-raise (CancelledError raw, others typed)
            end
        end
        deactivate Worker
        deactivate Session
        deactivate Caller
    end

    %% -- 3. Cancellation --
    rect rgb(0, 0, 0, 0)
        Note over Caller, Worker: Cancellation paths (wire-indistinguishable)

        alt Caller cancels (task.cancel / aclose / break)
            Caller ->> Connection: CancelledError at next await
            Connection ->> Connection: _call.cancel — RST_STREAM
            Connection ->> Service: gRPC stream cancelled
            Service ->> Session: async for raises CancelledError, except calls cancel()
        else Operator preempt (worker shutdown)
            Service ->> Session: _preempt iterates docket, calls cancel() on each session
        end
        Session ->> Worker: cancel worker task (call_soon_threadsafe)
        Worker ->> Worker: routine receives CancelledError, runs finally
        Worker -->> Session: _on_done resolves worker_done with CancelledError
        Session ->> Session: _iterate raises CancelledError at next yield
        Service -->> Connection: Response(exception=CancelledError)
        Connection -->> Caller: raise CancelledError
    end
```

### Unified worker dispatch driver

- Introduce `wool/runtime/worker/session.py` with `DispatchSession` and `Rejected` (internal parse-phase signal). One step is one request, one response. Coroutine routines synthesize a single `("next", None, None)` frame; streaming routines drive N frames against the same state.
- Collapse `WorkerService._run_on_worker` and `WorkerService._stream_from_worker` into a single dispatch entry that delegates to the session. The wool-task scope (`with task.context, with task, with do_dispatch(False)`) is established once per dispatch on the worker side regardless of routine shape.
- Schedule the worker driver lazily on first `__aiter__` so backpressure hooks run before any worker task acquires the context guard.
- Signal cancellation through a flag plus a direct cancel of the worker-loop task so a routine mid-step receives `CancelledError` instead of running to natural completion. `DispatchSession.cancel()` is the single mechanism for both caller-cancel and operator-preempt; the service-wide `WorkerService._preempt` iterates the docket and awaits `cancel()` on each session when forced cancellation is required. Parse-phase failures wrap into `Rejected` and ride the Nack-with-exception channel.
- Make `DispatchSession._iterate` raise `CancelledError` at the next yield boundary when `cancel()` has been signaled (both pre-iteration and in-flight), mirroring stdlib's `await task` semantics. Pre-fix, `cancel()` closed the response queue and `_iterate` exited cleanly — the dispatch handler's `async for` saw a clean termination and never shipped the terminal exception frame, leaving operator preempt silently invisible on the wire. Post-fix the iterator surfaces `CancelledError` directly, avoiding the race where the queue's pre-pushed `_EOS` arrives before the worker's `_on_done` callback could settle `worker_done` with the actual `CancelledError`.
- Register a `context.add_done_callback` on the gRPC servicer context that propagates client-side cancellation through `DispatchSession.cancel`. The naive reliance on `asyncio.Task` cancellation propagating through the dispatch generator's `await response_queue.get()` was unreliable on Python 3.11 + Linux; the done-callback gives a deterministic propagation signal across all supported Python versions.
- Extract `routine_scope` in `wool/runtime/routine/task.py` so caller-side `Task._run` / `Task._stream` and worker-side `DispatchSession` share routine-scope establishment. Async-generator `aclose` now matches stdlib semantics: routine-internal `CancelledError`, the synthesized `RuntimeError("async generator ignored GeneratorExit")` from a yield-during-GE protocol violation, and other exceptions propagate unchanged.
- Close the asyncio loop from the worker thread's `run` target so `run_forever`'s teardown owns it. Eliminates the `RuntimeError("Cannot close a running event loop")` race that fired when the caller's thread closed the loop in parallel with the worker thread's still-active `run_forever`.
- Record `StopRequest.timeout` on `_stop_timeout` so the protocol-provided budget governs both the docket-cancel and thread-join phases.

### Caller-side dispatch consolidation

- Collapse `WorkerConnection.dispatch`'s split-stack lifecycle into a single `AsyncExitStack` owned by the streaming generator. `_execute` is now the sole owner of every per-dispatch resource: the passthrough serializer ref (self-dispatch only), the channel ref, the concurrency permit, and the gRPC call's cancel hook. Acquire-then-register-release runs back-to-back with no intervening await, so cancellation cannot interleave between them.
- Drop the intermediate `_dispatch` method; extract the wire handshake as `_handshake` called from inside the stack.
- Drop the `channel` parameter on `_execute` and the idempotent-release closure that the previous design threaded through to cover the register-after-yield gap. Both are obsolete now that release is registered atomically with acquire.
- `WorkerConnection.dispatch`'s public contract is unchanged: same arguments, same primed-stream return, same exception classes.

### Wire-error classification

- Demote version-mismatch rejection from `Nack` to a gRPC `FAILED_PRECONDITION` abort emitted by `VersionInterceptor` before the request reaches the dispatch handler. Version handshake is transport-level concern, not parse-phase.
- Retain the `Nack`-with-exception channel for parse-phase rejections (malformed task id, unpicklable serializer hint, strict-mode `ContextDecodeWarning`, cloudpickle errors on the task callable, `ImportError`, non-async callable). The caller deserializes the original parse exception and re-raises it directly.
- Extend `wire.proto`'s `Nack` with an optional `Message exception` payload to carry the serialized parse exception. Plain Nacks fall back to `RpcError` when the payload is malformed.
- Switch `DispatchStream._read_next` and `_dispatch` from `except Exception` to `except BaseException` so `asyncio.CancelledError`, `KeyboardInterrupt`, and `SystemExit` mid-handshake or mid-read still cancel the in-flight gRPC call before re-raising.

### Cancellation propagation parity with stdlib

- Let routine-self-raised `asyncio.CancelledError` propagate raw out of `_read_next`. Stdlib `await task` is observationally indistinguishable between "task was externally cancelled" and "task self-raised `CancelledError`" — both transition the task to `CANCELLED` and the caller's `await` raises `CancelledError`. Wool preserves the same indistinguishability across the wire so caller code that catches `asyncio.CancelledError` keeps working.
- Degrade non-`CancelledError` `BaseException` payloads (`KeyboardInterrupt`, `SystemExit`, user-defined `BaseException` subclasses) to `UnexpectedResponse` instead of `RpcError`. `UnexpectedResponse` is not an `RpcError` subclass, so the load balancer treats it as a caller-fault and does not evict the worker — a routine-level fault no longer wipes a healthy worker from the pool.
- Caller cancellation (`task.cancel`, `aclose`, `async for ... break`) routes through `_call.cancel` → gRPC `RST_STREAM` → server-side `CancelledError` → `session.cancel` → worker-task cancel → routine receives `CancelledError` at its next suspension and runs its `try / finally`. Caller observes `CancelledError`. Worker-side `finally` runs.
- Operator-initiated preempt (worker stop) ships the worker-side `CancelledError` raw via the same routine-failure path as caller cancel. Wool's `await routine()` is wire-indistinguishable from stdlib's `await task` regardless of who initiated the cancel — operator, caller, or routine self-raise.

### Type-preserving exception serialization

- Make `_safely_serialize_exception` preserve the exception class through its fallback. Stdlib exception pickling round-trips `(type, args)` and drops `__traceback__`, `__cause__`, `__context__`, and `__notes__` — they're not part of the exception's identity. When a routine-level exception accumulates state that drags an un-picklable C-level object into the graph (e.g. a worker-thread frame on `__traceback__` reachable via `__cause__`), the first `cloudpickle.dumps` raises but the exception's *class* and *args* are still picklable on their own. Retry with `cls(*exc.args)` so the caller's `except RoutineError` keeps matching, mirroring the stdlib pickle contract for exceptions. Only demote to a synthesized `RuntimeError` when even the clean reconstruction fails to pickle — preserving the third-tier-always-succeeds guarantee.
- Extract `_attach_strict_mode_warnings` as a module-level helper so the strict-mode `ContextDecodeWarning` peer-attachment block sits in a flat sync function. Coverage tooling on Python 3.11 (`sys.settrace`) tracks the helper precisely; the previous deeply-nested form inside the dispatch async generator's nested except arm was opaque to pre-PEP-669 tracing and reported as uncovered despite running.
- Pre-fix, the fallback unconditionally synthesized a `RuntimeError(f"{type(exc).__name__}: {exc!s}")`, silently demoting any typed wool failure with un-picklable graph state to a generic exception class. Observable as a CI-only failure of `test_dispatch_ships_stop_async_iteration_raw_for_coroutine_routine` on Python 3.13.13/Linux/cloudpickle 3.1.2: the dispatch handler's PEP 525 unwrap correctly produced `StopAsyncIteration` but the serializer fallback shipped `RuntimeError` back to the caller. Stdlib parity is restored independent of which Python build triggers the underlying pickle failure.

### Load balancer dispatch contract

- Add a docstring to `LoadBalancerLike.dispatch` documenting the worker-health exception contract: `TransientRpcError` means skip-without-evict, non-transient `RpcError` means evict, any other class propagates without touching the pool.
- Bind the exception in `RoundRobinLoadBalancer.dispatch`'s handlers and log transient skips at `DEBUG` and non-transient evictions at `WARNING` with the worker uid attached.

### Documentation

- Replace the top-level README's `Error handling` section with a phase × source matrix covering caller-side request encoding, gRPC handshake, worker-side request decoding, routine execution, worker-side response encoding, caller-side response decoding, and post-execution teardown. Each phase specifies what wool-internal and user-code failures look like to the caller and what the load balancer does.
- Update the architecture sequence diagram to show `VersionInterceptor`'s pre-dispatch handshake check, the parse-phase Nack-with-exception path, lazy worker scheduling on first `__aiter__`, and the unified iteration flow.
- Rewrite the worker README's Task execution section around `DispatchSession`'s four phases (parse / iteration / teardown / cancellation). Add an Exception flow subsection covering the three worker-side exit channels.

### Misc

- Pin pyright `pythonVersion` to `3.11` so static analysis matches the runtime requirement (`BaseExceptionGroup`, `match`, `add_note`, `Task.cancelling` were being flagged as undefined under the default 3.7 assumption).
- Widen `wool_factory`'s `coro` parameter in `runtime/context/base.py` to `Coroutine | Generator` so the factory is a structural subtype of typeshed's `_TaskFactory` (which expects `_CoroutineLike[_T]` = `Coroutine | Generator`). Cast back to `Coroutine` inside the body where the existing code needs the narrower type. Removes the previous `# pyright: ignore` on `set_task_factory`.
- Add a unit test pinning `LocalDiscovery.Subscriber`'s rejection of negative `poll_interval`. The metaclass defers `__init__` to the resource-pool factory, so the validation fires on first iteration rather than construction — covering that surface with a positive test.

## Test cases

| 1 | `tests/runtime/worker/test_session.py` | A parsed `DispatchSession` for a coroutine routine | The session is iterated once | A single `_Response` is yielded carrying the return value and post-step context snapshot | Coroutine-shape driver path |
| 2 | `tests/runtime/worker/test_session.py` | A parsed `DispatchSession` for an async-generator routine | The session is iterated across multiple `("next", …)` commands | One response per yield, terminating with the generator's `StopAsyncIteration` | Streaming-shape driver path |
| 3 | `tests/runtime/worker/test_session.py` | A request with a malformed task id or unpicklable serializer hint | The session enters via `__aenter__` | `Rejected` is raised carrying the parse exception | Parse-phase rejection contract |
| 4 | `tests/runtime/worker/test_session.py` | A streaming session whose second `next` request's wire context decodes to a `BaseExceptionGroup` | The session pumps the second request | The error surfaces re-wrapped as "mid-stream request context decode failed" | Mid-stream decode-failure label |
| 5 | `tests/runtime/worker/test_session.py` | A session whose `__aenter__` parse path fails and whose `_stack.aclose` raises `KeyboardInterrupt` | The aclose helper runs | `KeyboardInterrupt` propagates raw, not swallowed under `except Exception` | Process-signal preservation during cleanup |
| 6 | `tests/runtime/worker/test_session.py` | A streaming session suspended at the request pump | `cancel` is called from another task before the next request lands | The pump observes `self._cancelled` at the top of its loop and the iterator raises `CancelledError` | Cancelled-mid-pump propagation |
| 7 | `tests/runtime/worker/test_session.py` | A `DispatchSession` whose iterator is suspended on `response_queue.get` | `cancel` is awaited from a different task | The suspended `anext` raises `CancelledError`, mirroring stdlib `await task` | Suspended-iteration cancel observable |
| 8 | `tests/runtime/worker/test_session.py` | A session whose worker loop has been torn down after the worker task was scheduled | `cancel` is awaited | `call_soon_threadsafe(worker_task.cancel)` raises and is swallowed; cancel returns cleanly | Cancel tolerance for closed worker loop |
| 9 | `tests/runtime/worker/test_service.py` | A `WorkerService` whose worker thread is running the loop | Loop teardown is requested | The loop closes from the worker thread inside `run_forever`'s finally without raising `RuntimeError("Cannot close a running event loop")` | Loop-close race fix |
| 10 | `tests/runtime/worker/test_service.py` | A `WorkerService` receiving a `StopRequest` with timeout=T | Stop is processed | `_stop_timeout` governs both docket-cancel and thread-join phases | StopRequest timeout plumbing |
| 11 | `tests/runtime/worker/test_service.py` | A `WorkerService` with an active coroutine task awaiting a long sleep | Stop RPC is called with timeout=0 | The routine observes `asyncio.CancelledError` within the test budget; the service signals stopped state and calls `proxy_pool.clear` | Operator-preempt routine observability (coroutine) |
| 12 | `tests/runtime/worker/test_service.py` | A `WorkerService` running an async-generator task mid-yield | Stop RPC is called with timeout=0 | The streaming routine observes `asyncio.CancelledError` or `GeneratorExit` within the test budget | Operator-preempt routine observability (streaming) |
| 13 | `tests/runtime/worker/test_service.py` | A coroutine routine that raises `StopAsyncIteration` at the top level | The dispatch handler ships the terminal-exception Response | The wire carries `StopAsyncIteration` raw — PEP 525's RuntimeError wrap is unwrapped for non-streaming routines | Stdlib `await coro()` parity |
| 14 | `tests/runtime/worker/test_service.py` | A `_safely_serialize_exception` call against an exception whose first `dumps` raises an un-picklable failure (`PickleError`, `TypeError`, `AttributeError`, `RecursionError`) | The fallback runs | The reconstructed `cls(*exc.args)` is shipped — the original exception class is preserved on the wire | Type-preserving fallback contract |
| 15 | `tests/runtime/worker/test_service.py` | A `_safely_serialize_exception` call whose reconstructed clean instance also fails to pickle | The tertiary fallback runs | A synthesized `RuntimeError` carrying the original class name and message is shipped — third tier always succeeds | Demotion tier (defense-in-depth) |
| 16 | `tests/runtime/worker/test_service.py` | A `_safely_serialize_exception` call whose first `dumps` raises a non-pickle failure (`ValueError`) | The helper runs | The unexpected failure propagates — the narrow catch refuses to silently absorb a serializer bug | Narrow-catch contract |
| 17 | `tests/runtime/worker/test_service.py` | A routine whose exception is un-picklable AND whose `__str__` raises | The dispatch handler ships the terminal exception | A synthesized `RuntimeError` carrying only the class name is shipped — the defensive `except Exception: message = cls_name` arm fires | Defensive message-construction fallback |
| 18 | `tests/runtime/worker/test_service.py` | A `WorkerService` whose `_stopping.is_set` returns `False` on the entry-gate check and `True` on the `_tracked` registration check | A dispatch request arrives | `_tracked` aborts with `UNAVAILABLE` and invokes `session.cancel` | Racy stop+dispatch arm |
| 19 | `tests/runtime/worker/test_interceptor.py` | A `VersionInterceptor` and a first request whose `TaskEnvelope` cannot be parsed | The interceptor runs | Abort with `FAILED_PRECONDITION` and a "Failed to parse version envelope" detail | Malformed envelope abort |
| 20 | `tests/runtime/worker/test_interceptor.py` | A first request whose version string is non-empty but unparseable | The interceptor runs | Abort with `FAILED_PRECONDITION` and a detail naming both client and worker versions | Unparseable-version abort |
| 21 | `tests/runtime/worker/test_interceptor.py` | A first request whose version is parseable but on a major incompatible with the worker (different major) | The interceptor runs | Abort with `FAILED_PRECONDITION` and a detail naming both versions | Incompatible-major abort |
| 22 | `tests/runtime/routine/test_task.py::TestRoutineScope` | A `routine_scope` block entered for a `Task` | The block runs and exits normally | Proxy is acquired, `wool.__proxy__` is bound, runtime context is entered, `do_dispatch` is toggled, and on exit each is reversed in stdlib order | Routine-scope contract |
| 23 | `tests/runtime/routine/test_task.py::TestRoutineScope` | A `routine_scope` async generator that yields after a `GeneratorExit` is thrown in | `aclose` is awaited | The synthesized `RuntimeError("async generator ignored GeneratorExit")` propagates per stdlib parity | aclose stdlib parity |
| 24 | `tests/runtime/worker/test_connection.py` | An in-flight `DispatchStream._read_next` | The task is cancelled mid-read | `self._call.cancel` runs before re-raising `CancelledError` | BaseException-scoped cleanup |
| 25 | `tests/runtime/worker/test_connection.py` | A `_execute` frame whose handshake-write fails after the channel-permit callback is registered | The frame unwinds | The channel semaphore is released via the exit-stack callback | Permit-leak fix |
| 26 | `tests/runtime/worker/test_connection.py` | A worker that rejects a dispatch during the parse phase | The client receives the Nack | The original parse exception (e.g., `ImportError`, `ContextDecodeWarning`) is deserialized and re-raised | Nack-with-exception channel |
| 27 | `tests/runtime/worker/test_connection.py` | A `Response.exception` carrying a non-Exception payload (a bare string) | The client iterates the response stream | `UnexpectedResponse` is raised — not `RpcError` — so the load balancer does not evict | Non-Exception payload degradation |
| 28 | `tests/runtime/worker/test_connection.py` | A `Response.exception` carrying an `asyncio.CancelledError` | The client iterates the response stream | `asyncio.CancelledError` propagates raw, not degraded to `RpcError` or `UnexpectedResponse` | Stdlib cancellation parity |
| 29 | `tests/runtime/worker/test_connection.py` | A `Response.exception` carrying a slotted exception that rejects `add_note` and arbitrary attribute writes, with strict-mode decode failures | The client iterates the response stream | The original exception type ships unchanged; the note/attribute writes are swallowed under the best-effort guards | Strict-mode note-attachment fallthrough |
| 30 | `tests/integration/test_unified_driver.py` | A real gRPC + worker-process boundary | A coroutine routine is dispatched (`await`) | The single-frame outcome is observed end to end | gRPC-level coroutine path |
| 31 | `tests/integration/test_unified_driver.py` | A real gRPC + worker-process boundary | A streaming routine is dispatched and consumed via `async for`, `asend`, `athrow`, `aclose` | Each verb yields the expected outcome per stdlib semantics | gRPC-level streaming verbs |
| 32 | `tests/integration/test_unified_driver.py` | The `ASYNC_GEN_ANEXT_SINGLE` scenario | A single-yield streaming routine is driven | The first iteration yields the value and the second terminates with `StopAsyncIteration` | Single-yield streaming edge case |
| 33 | `tests/integration/test_unified_driver.py` | A long-sleeping coroutine routine dispatched against an EPHEMERAL pool with a `try / finally` that writes a sentinel file | The caller wraps the dispatch in a task and cancels the caller task mid-sleep | `await task` raises `CancelledError` AND the sentinel file contains `"cancelled"` — worker-side `finally` ran | Caller-cancel propagates to worker |
| 34 | `tests/integration/test_unified_driver.py` | An async-generator routine yielding `"alive"` forever inside a `try / finally` writing a sentinel file | The caller advances once and calls `aclose` | The sentinel file contains `"cleaned_up"` — worker-side `finally` ran | Async-gen aclose triggers worker finally |
| 35 | `tests/integration/test_unified_driver.py` | A coroutine routine whose body is `raise asyncio.CancelledError(...)` | The caller awaits the routine | `asyncio.CancelledError` is raised on the caller, not `RpcError` or `UnexpectedResponse` | Self-raised cancellation surfaces as stdlib cancellation |
| 36 | `tests/runtime/discovery/test_local.py` | A `LocalDiscovery.Subscriber` constructed with a negative `poll_interval` | The caller starts iterating the subscriber (deferred-init triggers on first iteration) | `ValueError` is raised naming the bad value | Subscriber poll-interval validation |


wool/src/wool/runtime/routine/task.py
```diff
@@ -317,75 +320,67 @@ def to_protobuf(self, serializer: Serializer | None = None) -> protocol.Task:
...
+    ``with task.runtime_context, task, do_dispatch(False):`` once
+    around the caller's iteration. The routine (:class:`Coroutine` or
+    :class:`AsyncGenerator`) is built inside the scope and yielded
+    ready for stepping; on exit, async generators are ``aclose``'d
+    and coroutines are ``close``'d.
+
+    :data:`wool.__proxy_pool__` MUST be initialized before this is
+    entered — without a configured pool, nested dispatch fails
+    silently. The worker process initializes the pool at startup;
+    test scaffolding outside the worker must do the same.
```
conradbzura: This is pretty brittle, rewrite as more general prose that describes the behavior.

wool/src/wool/runtime/routine/task.py
```diff
@@ -317,75 +320,67 @@ def to_protobuf(self, serializer: Serializer | None = None) -> protocol.Task:
...
+    """
+    proxy_pool = wool.__proxy_pool__.get()
+    if proxy_pool is None:
+        raise RuntimeError("wool.__proxy_pool__ is not initialized")
+    # ``Task.__post_init__`` always seeds ``runtime_context`` from
+    # ``RuntimeContext.get_current()``; only a Task constructed
+    # without going through ``__post_init__`` (impossible in normal
+    # use) could leave it ``None``. The assertion narrows the type
+    # for pyright and surfaces any future shape regression loudly.
```
conradbzura: Drop this comment. The invariant is implied with the `assert` below.

wool/src/wool/runtime/routine/task.py
```diff
@@ -317,75 +320,67 @@ def to_protobuf(self, serializer: Serializer | None = None) -> protocol.Task:
...
+            with task.runtime_context, task, do_dispatch(False):
+                routine = task.callable(*task.args, **task.kwargs)
+                try:
+                    yield routine
+                finally:
+                    # ``aclose`` on async generators propagates the
+                    # routine's own exception (stdlib parity);
+                    # ``close`` on coroutines is a no-op on awaited
+                    # frames but prevents a "never awaited" warning
+                    # on the cancel-before-await path.
```
conradbzura: Drop this comment, code is self-explanatory.

wool/src/wool/runtime/worker/connection.py
conradbzura: Expand the `Exception` documentation with the comment here: https://github.com/wool-labs/wool/pull/205/changes#diff-46935f436e53471c594081c21df5fff3e6ac699b4effda69d20903dda4e4ea97R200-R210

Rephrase the comment to be appropriate for a docstring (i.e., make no reference to private symbols, etc.).

conradbzura: Didn't we refactor decide failures to ride on the primary exception on the worker side? This should match that behavior.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -376,6 +432,18 @@ class WorkerConnection:
...
+    an in-flight gRPC call wraps its body in
+    ``try / except BaseException`` so that ``asyncio.CancelledError``
+    (a :class:`BaseException` subclass, not :class:`Exception`) still
+    triggers ``call.cancel()`` before re-raising. The cancel itself
+    is swallowed at :class:`Exception` (not :class:`BaseException`) —
+    cleanup-during-cleanup should let ``KeyboardInterrupt`` propagate
+    rather than silently drop a process-level signal. The same shape
+    repeats in :meth:`_DispatchStream._read_next`,
+    :meth:`_DispatchStream.aclose`, :meth:`_dispatch`'s inner block,
+    and :meth:`_execute`'s tail.
```
conradbzura: Eliminate the brittle implementation-detail call sites, functions, etc.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -405,6 +473,15 @@ class WorkerConnection:
...
 
+    # gRPC status codes that classify as :class:`TransientRpcError`
+    # — worker is hiccupping; the load balancer should skip without
+    # eviction. Every other non-OK code (including ``INTERNAL`` for
+    # server-side bugs, ``FAILED_PRECONDITION`` for protocol skew,
+    # and the ``CANCELLED`` shape that ``context.abort`` raises)
+    # falls through to plain :class:`RpcError` so the LB evicts.
+    # Today's binary policy treats first-occurrence non-transient
+    # failures as eviction-worthy; health-aware forgiveness (an
+    # N-strike threshold) is a follow-up issue.
```
conradbzura: Drop this comment.

wool/src/wool/runtime/worker/connection.py
```diff
@@ -485,12 +562,23 @@ async def dispatch(
...
+            deserialized (malformed-dump fallback).
         :raises UnexpectedResponse:
             If the worker doesn't acknowledge the task.
         :raises ValueError:
             If the timeout value is not positive.
 
+        Parse-phase rejections (malformed task id, unpicklable
+        serializer hint, strict-mode :class:`wool.ContextDecodeWarning`,
+        cloudpickle errors on the task callable, ImportError on a
+        missing module, non-async callable) re-raise the original
```
conradbzura: Summarize the enumerated failure modes as "due to a malformed task payload".

wool/src/wool/runtime/worker/connection.py
```diff
@@ -591,6 +679,50 @@ async def _dispatch(
...
+                        # to Nack when the rejection originates from
+                        # parse-phase decode failures (malformed
+                        # task id, unpicklable serializer hint,
+                        # strict-mode ContextDecodeWarning,
+                        # cloudpickle errors on the task callable,
+                        # ImportError on a missing module on the
+                        # worker). Deserialize and re-raise so the
+                        # caller observes the actual failure class
+                        # rather than an opaque RpcError. Plain
+                        # Nacks (e.g., version mismatch) carry only
```
conradbzura: We should refactor the plain version-mismatch `Nack` as a gRPC error with a `FAILED_PRECONDITION` code. That way we don't have to wrap the `Nack` on the connection side.

PR #224, opened 2026-05-16
Reframe Wool's context model around stdlib contextvars — Closes #223, #229, #232

### LINKED ISSUE #223: Reframe Wool's context model around stdlib contextvars
## Description

Wool currently maintains a context system parallel to and fully disjoint from stdlib `contextvars`. `wool.Context` is a standalone container resolved through a task-keyed registry (`context_registry` in `wool/runtime/context/registry.py`), and its own docstring states that a `wool.Context` and a `contextvars.Context` "never interact". Propagation is therefore limited to the asyncio task tree, via the task factory `WorkerPool` installs. Every non-task scheduling edge — `loop.call_soon`, `loop.add_reader` callbacks, `Future.add_done_callback`, `asyncio.to_thread` — falls through to a single flat per-thread fallback `Context`, so a callback neither inherits the scheduling scope's Wool state nor is isolated from other callbacks on the same thread. The model also cannot reach callbacks at all under uvloop, whose I/O scheduling is internal to libuv and exposes no interceptable seam.

This refactor re-founds the context model directly on stdlib `contextvars`. Wool context state rides in stdlib context variables; `wool.ContextVar` becomes a `contextvars`-backed variable that additionally (a) propagates across worker boundaries and (b) opts its chain into Wool's linearity rules. The standalone `wool.Context` type is removed, and with it `current_context()`, `copy_context()`, and `wool.create_task()` — the last of which exists today purely as a typing shim, because `wool.Context` cannot subclass the C-level `contextvars.Context`. `wool.ContextVar` becomes the sole public context-variable API.

Wool's invariants are still enforced, on top of stdlib propagation: forced copy-on-fork (every task fork receives a fresh chain UUID, minted by the task factory) and strictly prohibited concurrent entry (a guard that fails loud when a chain would be run by two runners at once — a second OS thread, or a second concurrently-live task handed the same context). Enforcement is "armed-gated": it engages only once a `wool.ContextVar` has actually been set on a context. An unarmed context — one carrying no Wool variables — behaves as a plain stdlib `contextvars.Context`, so Wool imposes nothing on code that does not use it.

Divergence from current behavior is accepted in pursuit of full stdlib parity. A tighter coupling between `contextvars` and Wool is unavoidable and intended: snapshotting Wool state now means snapshotting the `contextvars` context, and isolating the `contextvars` context now isolates Wool state with it.

## Motivation

- The current model cannot propagate Wool context to event-loop callbacks, and cannot reach callbacks at all under uvloop — the default event loop for `uvicorn[standard]` and a staple of production asyncio deployments. Any integration that spawns work from a non-task scheduling edge (notably an ASGI/Uvicorn integration, where request tasks are created inside an I/O callback) cannot rely on Wool context reaching that work.
- Loop interception (a custom event loop, or monkeypatched scheduling methods) was evaluated and rejected: it is structurally limited to the pure-Python `SelectorEventLoop` and cannot reach uvloop. Riding on `contextvars` is the only propagation vehicle every conformant loop, uvloop included, is obligated to honor.
- Maintaining a second, parallel context system is a standing source of confusion. A single context system with one clearly-marked special variable type is a materially simpler mental model than reconciling two systems that "never interact".
- The per-thread fallback is not merely "no inheritance" — it is a single shared `Context` for every non-task scope on a thread, so unrelated callbacks silently share Wool state. Removing it removes a latent correctness hazard.

## Storage model

Wool chain state is held as a single immutable snapshot in one wool-owned stdlib `contextvars.ContextVar`. The snapshot carries the chain UUID, the map of `wool.ContextVar` key to value, and the consumed-token log. `wool.ContextVar.get` reads its key from the snapshot; `set` and `reset` rebuild the snapshot and reinstall it. The snapshot's immutability is what makes sharing it by reference across callbacks safe without intercepting the event loop — a callback that sets a variable produces a new snapshot scoped to its own `contextvars` copy, leaving the scheduling scope's snapshot untouched.

The considered alternative — backing each `wool.ContextVar` with its own stdlib `contextvars.ContextVar` — is rejected, primarily because of cross-process reset propagation. When a routine resets a variable on a worker, the caller-side response merge must *remove* that variable from its context. `contextvars.ContextVar` has no delete primitive: the only way to un-set one is `reset(token)`, with a token held from a prior `set`. A variable set by an ancestor of the caller has no caller-side token, so per-variable backing cannot remove it on merge without encoding "unset" as a magic sentinel that every `get` must then special-case. A single owned snapshot removes the key simply by rebuilding the map. Per-variable backing also cannot fully delegate to stdlib regardless: a stdlib `reset` is invisible to Wool and cannot be serialized, so a wool-owned reset log is required either way. The single-snapshot model is additionally the smaller, lower-risk delta — it re-homes today's `Context._data` map and `wool.Token` / consumed-token machinery from a task-keyed object into a `contextvars`-held snapshot, rather than rewriting that layer.

## Expected Outcome

1. `wool.ContextVar` values are stored in stdlib `contextvars` and propagate with stdlib visibility across every conformant event loop — uvloop included — and across every asyncio scheduling edge: task creation, `call_soon`/`call_later`/`call_at`, `add_reader`/`add_writer`/`add_signal_handler`, and `Future.add_done_callback`.
2. The public `wool.Context` type is removed, along with `current_context()`, `copy_context()`, `wool.create_task()`, and `ContextAlreadyBound`. `wool.ContextVar` is the sole public context-variable type; two new public symbols join the context surface — `wool.to_thread` (the Wool-aware thread-offload fork) and `wool.ConcurrentChainEntry` (the concurrent-entry guard exception). `RuntimeContext` is unaffected. The `__all__` of `wool/__init__.py` is updated accordingly. This is a breaking change to the public API; backward compatibility is explicitly not preserved.
3. Wool's invariants hold under the new model: every task fork receives a fresh chain UUID (copy-on-fork via the task factory), and concurrent entry into a single chain fails loud with `ConcurrentChainEntry`. The guard is not keyed on the chain UUID; it has two enforcement points — (a) a `wool.ContextVar` access (`get`/`set`/`reset`) from a thread other than the chain's owner thread raises, catching an armed chain reached from an executor thread; (b) the task factory raises up front when one `contextvars.Context` object is handed to `create_task` for two concurrently-live tasks.
4. Enforcement is armed-gated. A `contextvars` context carrying no `wool.ContextVar` behaves as a plain stdlib context: no chain UUID, no guard, no fork bookkeeping, and no measurable overhead for code that does not use Wool.
5. Cross-worker dispatch continues to propagate Wool-variable state and chain identity over the wire protocol. The `Context.to_protobuf` / `from_protobuf` / `update` methods become internal `encode_snapshot` / `decode_snapshot` / `merge_snapshot` functions operating on the current `contextvars` context, not methods on a public type.
6. Cross-process reset propagation — the consumed-token semantics that today carry a "this variable was reset" signal across the wire — is preserved.
7. Offloading to another OS thread from an armed context never silently shares a chain across true parallelism: `wool.to_thread` forks explicitly into a fresh, detached chain (no merge-back), while a plain `asyncio.to_thread` from an armed context fails loud.
8. The stdlib parity matrix below holds. Each touch point reaches functional parity, restricted only where Wool's parallelism genuinely requires it.
9. `tests/stdlib_parity/` is expanded to assert the matrix — propagation into tasks, callbacks, timers, I/O-readiness callbacks, and done-callbacks, plus the armed-gating and guard behaviors.

## stdlib parity matrix

The `contextvars` API surface:

| Touch point | Parity approach | Restriction from Wool parallelism |
|---|---|---|
| `ContextVar(name, *, default)` | `wool.ContextVar` mirrors the constructor, adding an optional `namespace`. | Keys must be unique within a namespace (`ContextVarCollision`); values that cross workers must be cloudpicklable. |
| `ContextVar.get` / `set` / `reset` | Operate on the wool-owned snapshot with stdlib semantics; `set` returns a `Token`. | The first `set` on a context arms it — mints the chain UUID and engages the guard. |
| `contextvars.Token` | Wool retains a wire-serializable `Token` so resets propagate across workers. | `reset` is rejected if the token was minted in a different chain (existing behavior). |
| `contextvars.Context` (the type) | Wool provides no replacement; callers hold stdlib `contextvars.Context`. Wool variables ride in the wool-owned snapshot variable within it. | An armed stdlib context is subject to the concurrent-entry guard. |
| `Context.run` | stdlib `Context.run`, unchanged. | Running an armed context from two places at once trips the guard. |
| `Context.copy()` / `copy_context()` | stdlib, unchanged. | A bare copy keeps the chain UUID (same chain); a fork with a fresh UUID happens only through task creation. Running a copy concurrently with its origin trips the guard. |

asyncio propagation touch points:

| Touch point | Parity approach | Restriction from Wool parallelism |
|---|---|---|
| `Task` creation (`create_task`, `ensure_future`, `TaskGroup`) | The Wool task factory forks: the child inherits the parent's snapshot and receives a freshly minted chain UUID. | None beyond the fresh UUID, which is correct fork semantics. Dormant when unarmed. |
| `call_soon` / `call_soon_threadsafe` | `copy_context()` carries the snapshot variable into the `Handle`; the callback runs with full visibility of the scheduling scope's Wool variables. | The callback shares the scheduling scope's chain UUID. Safe — callbacks are synchronous and cooperatively scheduled, so they never run concurrently with their scheduler; the guard arbitrates tasks only. |
| `call_later` / `call_at` | Same as `call_soon` — the `TimerHandle` carries the copied context. | Same as `call_soon`. |
| `add_reader` / `add_writer` / `add_signal_handler` | `copy_context()` is captured once at registration; the callback sees the snapshot as of that point. | The snapshot is captured at registration and reused for every readiness event — inherited stdlib behavior, not Wool-specific. |
| `Future.add_done_callback` | `copy_context()` at registration carries the snapshot into the done-callback. | None — the done-callback is synchronous. |
| `loop.run_in_executor` (bare) | stdlib does not copy context into the executor thread; Wool matches — bare `run_in_executor` does not carry Wool variables. | To carry context into an executor, use the `to_thread` path below. |
| `asyncio.to_thread` | The offloaded thread runs in genuine parallelism with the loop. Wool provides an explicit thread-offload fork that mints a fresh chain UUID and ships a detached snapshot with no merge-back. | Plain `asyncio.to_thread` from an armed context would place a second runner on the same chain in parallel — this fails loud, directing the caller to the Wool-aware fork. This is the single divergence-with-restriction. |
| `loop.set_task_factory` / `get_task_factory` | Wool installs its factory here and composes with any existing factory. | The "install last" ordering contract holds: a user factory installed after Wool's drops the fork wrapping. |

## Out of scope

- Migrating `__proxy__` / `__proxy_pool__` to `wool.ContextVar`, and the Wool/Uvicorn ASGI integration, build on this refactor but are tracked separately.
- A cross-process concurrent-entry guard remains out of scope: the per-process guard neither solves nor regresses it (today's per-object guard does not survive the wire either).

### LINKED ISSUE #229: Split the decode-side manifest transient into a dedicated type
## Description

Factor the wire-decoded context transient out of the live context model into a dedicated type. The live model should be unambiguously live — never a half-decoded variant — and the decoded shape should be its own wire-view value object.

Split the model in two:

- The live chain (`Chain`) carries only its bound-var index: `id`, `vars`, and `resets`. It holds no decode-side `_manifest` field, so there is no decoded-but-not-live variant of it.
- A dedicated manifest type (`_ChainManifest`) carries the deserialised state of a wire `ChainManifest`: `id`, `vars`, `resets`, `stubs`, and `decode_error`. It is a wire-view POD — never installed, never the active chain state.

The decode path returns the manifest type rather than a half-populated chain. A decoded manifest lives on the dispatch session and the frame (`DispatchSession.decoded`, `Frame.context`); `Frame.mount` merges it into the live chain key-wise at mount time.

Leave `external_used` out of the manifest. Cross-process token transport — the consumed-token log's only consumer — is out of scope here and tracked in #231.

## Motivation

`Context` carries three near-overlapping concepts in one class: the bound-var index, the decode-side `dict[var, value]` transient (`_manifest`), and a `vars` proxy over it. "Bound" lives in one field; "value" lives elsewhere — except in the window between decode and mount, when it lives in `_manifest`. A reader has to hold that exception in their head to follow the model. A dedicated decoded type removes it: the live chain is always live, and the decoded shape is a separate thing with a separate name.

## Expected Outcome

- `Chain` has no `_manifest` field and is always a live chain.
- A dedicated manifest type holds the wire-decoded state (`id`, `vars`, `resets`, `stubs`, `decode_error`) and is never installed as active state.
- The decode path yields the manifest type; `Frame.mount` merges it into the live chain key-wise.
- The manifest carries no `external_used` log; cross-process token transport stays out of scope (#231).

### LINKED ISSUE #232: Add IneffectiveLeaseWarning and IneffectiveQuorumTimeoutWarning to wool.__all__
## Description

Re-export `IneffectiveLeaseWarning` (defined in `wool.runtime.worker.pool`) and `IneffectiveQuorumTimeoutWarning` (defined in `wool.runtime.worker.proxy`) from `wool/__init__.py`, and add both names to `wool.__all__`. Place them on the public package surface beside `SerializationWarning`, under the `WoolWarning` umbrella.

## Motivation

Both warnings are user-tunable through the standard warnings filter — `warnings.filterwarnings("error", category=...)` or `PYTHONWARNINGS=error::wool.IneffectiveLeaseWarning` — but neither has a stable public import path. A caller who wants to silence, promote, or filter one must reach into a private module: `from wool.runtime.worker.pool import IneffectiveLeaseWarning`. That couples application code to Wool's internal layout. `SerializationWarning` already sits on the package surface; these two belong there too.

## Expected Outcome

- `from wool import IneffectiveLeaseWarning, IneffectiveQuorumTimeoutWarning` resolves.
- Both names appear in `wool.__all__`, beside `SerializationWarning`, and the public-API allowlist test covers them.
- The emission sites in `pool.py` and `proxy.py` are untouched — no behavior change.

conradbzura:
## Summary

Re-found Wool's context model directly on Python's stdlib `contextvars`. There is one context system, not two: Wool chain state rides in a single Wool-owned stdlib `contextvars.ContextVar` (`wool.__chain__`) as an immutable `Chain` dataclass, so it propagates with stdlib visibility across every conformant event loop (including uvloop) and across every cooperative asyncio scheduling edge — `call_soon` / `call_later` / `call_at`, `add_reader` / `add_writer` / `add_signal_handler`, `Future.add_done_callback`. Child task creation propagates as a *fork* — the child inherits the parent's variable values on a freshly minted chain ID, not the parent's chain itself. OS-thread offload is the one deliberate exception: an armed chain entered from a second OS thread fails loud rather than propagating silently. The chain-contention guard arms only when a `wool.ContextVar` has been set; a process that touches neither `wool.ContextVar` nor the dispatch path pays nothing. Entering a `WorkerProxy` (or a pool) counts as such a touch: `WorkerProxy.__aenter__` arms a guarded `_armed` marker so two tasks sharing one `contextvars.Context` cannot silently clobber each other's `wool.__proxy__`, so dispatching through a proxy always arms the chain — even when the routine sets no `wool.ContextVar` of its own.

Reshape the public API surface. Remove `wool.Context`, `wool.current_context`, `wool.copy_context`, `wool.create_task`, and `wool.ContextAlreadyBound`. Alias `wool.Token` to stdlib `contextvars.Token` (`wool.Token is contextvars.Token`); user code reuses the stdlib implementation directly and the previous home-grown Token class is deleted. Add `wool.to_thread`, `wool.install_task_factory`, and an exception hierarchy rooted at two umbrellas — `wool.WoolError` and `wool.WoolWarning` — with `wool.ChainContention`, `wool.ContextVarCollision`, `wool.TaskFactoryDisplaced`, `wool.SerializationError`, `wool.ChainSerializationError`, and `wool.SerializationWarning` underneath. `wool.ContextVar` and `wool.RuntimeContext` carry over.

Introduce a Frame/ChainManifest wire-frame abstraction in `wool.runtime.worker.frame`: every request and response on the worker wire wraps in a `Frame` subclass (`TaskRequestFrame`, `SendRequestFrame`, `ThrowRequestFrame`, `NextRequestFrame`, `ResultResponseFrame`, `ExceptionResponseFrame`, `AckResponseFrame`, `NackResponseFrame`) that carries its own decoded sender-side state alongside the payload. Receive sites mount that state into the active chain as a single canonical step, replacing the ad-hoc wire-context plumbing previously scattered across `worker/session.py`, `worker/connection.py`, and `worker/service.py`.

This is a breaking change to the public API; backward compatibility is not preserved.

Closes #223, #229, #232

The diagram below traces one `wool.ContextVar` value across a cross-process routine dispatch — arm, encode, apply, mutate, and apply back:

```mermaid
sequenceDiagram
    participant Caller
    participant Connection
    participant DispatchSession
    participant Routine

    rect rgb(0, 0, 0, 0)
        Note over Caller,Routine: Arm & dispatch
        Caller->>Caller: async with WorkerProxy / pool — WorkerProxy._armed.set(True) mints the chain ID and arms the Chain
        Caller->>Caller: wool.ContextVar.set(v) — bind v on the already-armed chain
        Caller->>Connection: dispatch routine
        Connection->>DispatchSession: TaskRequestFrame — pure dispatch metadata, no chain manifest
        Connection->>Connection: NextRequestFrame.for_send — capture the armed chain (_armed + v) into a ChainManifest
        Connection->>DispatchSession: NextRequestFrame
    end

    rect rgb(0, 0, 0, 0)
        Note over Caller,Routine: Worker apply & run
        DispatchSession->>DispatchSession: Frame.from_protobuf — decode, then mount the ChainManifest into the worker's cached context
        DispatchSession->>Routine: drive (await / asend)
        Routine->>Routine: wool.ContextVar.get() reads the caller's v
        Routine->>Routine: wool.ContextVar.set(v') mutates on the chain
    end

    rect rgb(0, 0, 0, 0)
        Note over Caller,Routine: Apply back
        Routine-->>DispatchSession: result
        DispatchSession->>DispatchSession: ResultResponseFrame.for_send — auto-capture the worker chain into a ChainManifest
        DispatchSession-->>Connection: ResultResponseFrame
        Connection->>Connection: Frame.from_protobuf — caller adopts worker mutations
        Connection-->>Caller: result — wool.ContextVar now reads v'
    end

    Note over Caller,Routine: The worker caches one contextvars.Context per chain ID — async-generator routines repeat apply/run and apply-back once per __anext__ / asend via NextRequestFrame / SendRequestFrame / ResultResponseFrame, reusing that cached context across yields
    Note over Caller,Routine: Entering a proxy arms the chain (WorkerProxy._armed), so a routine dispatched through a proxy always propagates an armed chain — even one that sets no wool.ContextVar of its own
```

## Design rationale

**One context system, not two.** Wool's previous design ran a second context system beside the standard library's: a `wool.Context` value store keyed by asyncio task, its own chain UUID, its own copy-on-fork task factory, a home-grown `Token`, and a registry that mirrored `contextvars` task-by-task — roughly 1,400 lines across `base.py`, `stub.py`, and `token.py` whose own docstring conceded that a `wool.Context` and a `contextvars.Context` "never interact." That parallel system existed only because Wool kept state outside `contextvars`. Re-founding state on a single Wool-owned `contextvars.ContextVar` (`wool.__chain__`), with each variable's value living in its own backing `contextvars.ContextVar`, obviates much of the prior design. The standard library already provides scoping, copy-on-`create_task`, value sharing across an `await`, and propagation across every conformant loop (uvloop included) and every cooperative scheduling edge. The tell that the old machinery was redundant: `RuntimeContext` and its `dispatch_timeout` were already plain `contextvars` and needed none of it, so they survived the teardown essentially unchanged.

> **Alternative considered: refactor `base.py` in place.** The 953-line monolith was removed wholesale rather than refactored because the new model is a different substrate, not a cleaner version of the old one. Every verb the old surface owned — `current_context`, `copy_context`, `create_task` — collapses to `contextvars.copy_context` or `asyncio.create_task`, so a refactor would have preserved an unnecessary abstraction.

**The chain is an index, not a value store.** The `Chain` is a frozen, identity-equality dataclass that records which variables are bound, the logical-chain UUID, the owner stamps, the reset signals, and the stub pins. Each variable's value lives in its own backing `contextvars.ContextVar`. The governing invariant is that `X` is in `chain.vars` exactly when `X`'s backing resolves in the active context, so all value storage, copy-into-child-tasks, copy-into-callbacks, and "writes in a copy never leak back" isolation are inherited from `contextvars` rather than reimplemented. Storing values inline on the `Chain` would duplicate and fight the standard library's copy-on-write and break the leak isolation the model rests on. Every mutation goes through `Chain.mount`, the single arming boundary that re-stamps ownership, rewinds resets, sets `wool.__chain__`, and ensures Wool's task factory is installed. The ensures copy-on-fork discipline is unbypassable and a displaced factory is caught at the same call.

**`wool.ContextVar` is a thin behavioral layer; `wool.Token` is the standard library's token.** Each `wool.ContextVar` wraps a private `contextvars.ContextVar` and delegates `get` / `set` / `reset` to it, and `wool.Token is contextvars.Token`. The standard library then enforces single-use, wrong-variable rejection, and wrong-context rejection itself, atomically: `reset` runs the native reset first, so if it raises, no Wool bookkeeping has happened, and cross-task isolation falls out for free because the standard library rejects a token minted in a different `contextvars.Context` and a fork runs in a copy. Variable identity stays the `(namespace, name)` key, and a second declaration of one key raises `ContextVarCollision`. Deleting the 297-line home-grown `Token` is what this buys. The old token carried a wire UUID, the prior value, the owning chain's UUID, a used flag, and a reconstitution apparatus whose entire job was to ship reset and used state across the wire; once a reset is just an absent-value entry in the chain manifest, the token never crosses the wire and collapses to the pure process-local revert handle the standard library already ships.

**Enforcement is armed-gated, so a process that never sets a Wool variable pays nothing.** While `wool.__chain__` is unset it defaults to `None`: the surrounding context is unarmed and behaves as a plain `contextvars.Context`, with no chain UUID and no guard. Two events arm it — the first `wool.ContextVar.set()`, or an incoming wire manifest; entering a `WorkerProxy` is a third in practice, because `WorkerProxy.__aenter__` arms a guarded `_armed` marker so concurrent tasks cannot clobber each other's `wool.__proxy__`, so a proxy-routed dispatch always arms even when the routine sets nothing of its own. Arming is deferred to first use rather than installed eagerly so that copying an untouched context carries no extra variable and incurs no contention overhead, and so Wool composes cleanly with `contextvars` for code that never opts in.

**Copy-on-fork is minted actively, not inherited from asyncio.** Child task creation propagates as a fork: the child inherits the parent's variable values on a freshly minted chain UUID, installed by Wool's task factory wrapping the child coroutine and re-owning a fresh chain inside the child's own running context. asyncio's native `copy_context()` duplicates the bindings, but the duplicate still references the same `Chain` object, so parent and child would read and write one shared chain. The invariant that a child never shares its parent's chain requires actively minting a fresh ID, which only code running under the child's own context can do.

> **Alternative considered: replace the loop's task factory.** `install_task_factory` composes over any existing factory through a `__wool_inner__` link instead of replacing it, and the composition walk detects Wool even when buried as `wool → third-party → wool`. Replacing outright is hostile to other libraries, and an outer-only idempotency check would miss a buried prior install and re-wrap into a double fork. Because copy-on-fork silently lost is a latent correctness bug, displacement by a later third-party factory is made loud rather than prevented: `TaskFactoryDisplaced` fires from three converging detection points — a `weakref.finalize` on Wool's factory, a per-task done-callback backstop, and the factory check that every chain mount runs — so even a worker surfaces displacement at frame ingress.

**One chain, one runner invariant.** The chain-contention guard enforces a single invariant: a chain is never held by two runners (i.e., a thread, task, or callback) at once. Entering an armed chain from a thread other than its owner or handing a context that is already driving a live task to a second `create_task`raises `ChainContention`. Admitting the second runner silently is the last-write-wins corruption the invariant exists to prevent, so the guard raises at the `wool.ContextVar` access where the violation first becomes observable. Cooperatively-scheduled work is deliberately not arbitrated: `call_soon` / `call_later` / `call_at`, `add_reader` / `add_writer` / `add_signal_handler`, and `Future.add_done_callback` run synchronously on the owning thread, so they share the scheduling scope's chain. The exception carries structured, picklable diagnostics — `chain_id`, a `kind` of `"thread"` or `"task"`, and the offending thread ID or task.

**OS-thread offload is the one deliberate exception to silent propagation.** An armed chain is single-owner, so two OS threads genuinely running in parallel cannot share it. `wool.to_thread` is the supported offload: it forks the caller's bindings onto a freshly minted chain owned by the new thread, with no back-propagation. The fork is performed on the new thread, not the calling thread, so the owner stamp lands on the thread that will run it. A plain `asyncio.to_thread` from an armed context is left to fail loud rather than patched to copy the context silently, as the latter may be surprising to users — users must explicitly opt into Wool's context system.

**Decode never preempts the routine's primary signal.** A routine's result or raised exception is the primary outcome of a dispatch; the propagated chain snapshot riding alongside it is ancillary. The wire snapshot is a distinct `ChainManifest` — decoded but unmounted, never touching a live `contextvars.Context` — so `ChainManifest.from_protobuf` is a pure decoder that can fail without raising. A `Frame` carries the decode result as a `ChainManifest | ChainSerializationError | None` union and defers the failure to mount time, where the payload decides how it surfaces: a value frame raises it, but an exception frame walks the payload's `__context__` chain and attaches the decode error at the tail, so a malformed manifest never displaces the exception the caller is waiting to catch. Per-entry failures degrade to a `SerializationWarning` and skip the entry; strict mode (`PYTHONWARNINGS=error::wool.SerializationWarning`) promotes and aggregates them into one `ChainSerializationError` after the loop, so every bad entry surfaces at once. A value-decode failure drops its key from both the bindings and the resets, so a half-decoded entry can never read back as a phantom reset that pops a live binding on the receiver.

**Each frame owns its own serialization.** Every request and response on the worker wire is wrapped in a `Frame` subclass that encodes its payload, captures or accepts a chain manifest, and mounts that manifest at receive time as one canonical step. The prior design scattered wire-context application across `session.py`, `connection.py`, and `service.py`, which is the seam where caller-side and worker-side encodings could drift apart; concentrating it in one class per wire variant removes that seam, and a subclass registry makes decode a total lookup over the payload that fails loudly on schema drift rather than silently. Boundary frames — the `TaskRequestFrame` dispatch handshake and the `Ack` / `Nack` responses — carry no manifest; the chain rides only on the mid-stream drive frames (`Next` / `Send` / `Throw` / `Result` / `Exception`). The armed chain therefore propagates with the first drive frame, captured at the moment of the first pull, not with the dispatch metadata that is acknowledged before the routine body runs.

**The worker caches one `contextvars.Context` per chain ID.** The dispatch driver keys a `WeakValueDictionary[UUID, contextvars.Context]` by wool chain ID. Frames sharing a chain ID reuse one context, so a routine's `set` before an async-generator yield and its `reset` after land in the same context — the standard library drives every resumption of one generator in a single context, and this reproduces that parity across the process boundary. Frames with distinct chain ids resolve to distinct contexts and stay isolated. Each step runs in a task built directly from the `asyncio.Task` constructor, bypassing the installed fork factory, because an internal driver must drive an existing chain rather than fork it onto a new ID that would break the very lookup the next frame depends on. The driver mounts task-agnostically — the cached context is driven by a succession of step-tasks, so it is owned by its thread, but never pinned to one task. An unarmed dispatch runs its frames in a single working context left unarmed, so a stateless routine's plain `asyncio.to_thread` offload copies a bare context and never trips the guard; if the routine arms that context with its own `set`, the driver re-indexes it under the freshly minted chain ID so the next back-propagated frame reuses it.

**The wire message is value-only.** The wire message is renamed from `Context` to `ChainManifest`, and each variable entry carries only its value, with an absent value standing as the entire reset signal. The old model needed a repeated `consumed_tokens` ledger — hex token ids propagated in both directions so a receiver could recognise a minted-then-reset token and pop the matching variable. The chain manifest is self-describing: an unset value already means "reset to no prior value on the sender's chain," so the ledger and its double-reset bookkeeping are deleted outright, and the wire shrinks to a snapshot of the sender's current bindings.

**The public surface is the standard library plus thin wrappers, with no compatibility shims.** `wool.Context`, `wool.current_context`, `wool.copy_context`, `wool.create_task`, and `wool.ContextAlreadyBound` are removed; `wool.Token` aliases `contextvars.Token`; `wool.to_thread` and `wool.install_task_factory` are added. The removed verbs are replaced by `contextvars.copy_context` and `asyncio.create_task`, and keeping aliases would perpetuate the parallel vocabulary the reframe intends to retire.

### Chain-backed storage

Wool chain state — the bound-`wool.ContextVar` index, the logical-chain UUID, the owner-thread / owning-task ownership stamps, the reset-variable signals, and the stub-pin set — rides in a single immutable `Chain` dataclass (`runtime/context/chain.py`, frozen, identity-equality) held in one Wool-owned stdlib `contextvars.ContextVar` (`wool.__chain__`). The `Chain` is an *index*, not a value store: each `wool.ContextVar`'s value lives in its own backing `contextvars.ContextVar`, and the chain's `vars` field records which variables are bound. Because the values live in the stdlib `contextvars.Context` rather than in the chain, callback write-isolation rests on native `contextvars` copy-on-write: a callback runs in a `copy_context()` copy, so its backing-variable writes (and the new chain it installs) stay in that copy.

`Chain.mount(*, owned=True)` is the single arming boundary: it re-stamps the owning thread — and, on an owned mount, the owning task — from the calling scope, rewinds reset signals, sets `wool.__chain__`, and unconditionally ensures Wool's task factory is on the running loop (a no-op once installed; the same call that detects factory displacement). Wire ingress goes through `Chain.from_manifest(manifest, *, owned, merge_with=None)`, entered from `Frame.mount`: it drains the decoded values into their backing variables, then either seeds a fresh chain from the manifest alone (`merge_with=None`) or unions the manifest onto a live receiver chain that keeps its chain ID (`merge_with` set), and delegates the arming to `Chain.mount`. The worker per-step driver mounts task-agnostically (`owned=False`): its cached `contextvars.Context` is driven by a succession of step-tasks, so the chain is owned thread-wise but never pinned to one task. Incoming wire state wins on overlap; reset-only keys drop out so the reset propagates.

Reorganise the context module around the new model. The live `Chain` lives in `chain.py`; the stdlib-compatible `ContextVar` (backed by a per-var `contextvars.ContextVar`) in `var.py`; the chain-ownership guard in `guard.py`; a shared, cycle-free `ContextVar` registry in `registry.py`; `install_task_factory` and the task-factory composition path in `factory.py`; the stdlib-aligned `to_thread` in `threading.py`; and the `RuntimeContext` overlay for dispatch-time fields in `runtime.py`. Deferred wire-side state is decoupled from the live chain: `ChainManifest` (the decoded-but-unmounted snapshot) lives in `manifest.py`, and the context error hierarchy is centralised in `runtime/context/exceptions.py`.

### Reshape the public API and root the exception hierarchy

Remove `wool.Context`, `wool.current_context`, `wool.copy_context`, `wool.create_task`, and `wool.ContextAlreadyBound`. Add `wool.to_thread` and `wool.install_task_factory`. `wool.ContextVar` and `wool.RuntimeContext` carry over unchanged.

Root the exception surface at two umbrellas: `wool.WoolError(Exception)` and `wool.WoolWarning(Warning)`, both in `src/wool/exceptions.py`. Under them: `wool.SerializationError`, `wool.ChainSerializationError` (a `SerializationError`), `wool.SerializationWarning`, `wool.ChainContention`, `wool.ContextVarCollision`, and `wool.TaskFactoryDisplaced`, all defined in `runtime/context/exceptions.py`. The earlier `ContextSerializationError` / `ContextDecodeError` / `ContextDecodeWarning` names are removed outright — no back-compat aliases.

Alias `wool.Token` to stdlib `contextvars.Token` — `wool.Token is contextvars.Token`, with no Wool subclass or shim. A token is therefore a plain stdlib token: it carries no Wool-internal state and does not transport across the wire (vanilla `pickle` / `cloudpickle` / `copy` raise `TypeError`), so value-state propagation rides the chain manifest rather than the token. The dispatch-serialisation machinery lives on `wool.ContextVar` (guarded against vanilla pickling), not on the token.

### Armed-gating, copy-on-fork, and the chain-contention guard

Enforcement is armed-gated. While unset, `wool.__chain__` defaults to `None`; the surrounding context is *unarmed* and behaves as a plain `contextvars.Context` — no chain UUID and no guard. A context is armed by either of two events: the first `wool.ContextVar.set()` on it, or a `Chain.from_manifest(...)` applying incoming wire state.

The chain-contention guard enforces one invariant — a single chain is never run by two runners at once — across both trigger paths:

- **No shared context across concurrent tasks.** The task factory forks every child task onto a freshly minted chain UUID (copy-on-fork), so sibling tasks created the ordinary way never share a chain. The guard bites when a `contextvars.Context` is explicitly reused: handing a context that is already driving a live task to a second `create_task` raises `ChainContention`. Once the first task completes, that context is free to drive another.
- **No cross-thread entry.** Entering an armed chain from a thread other than its owner raises `ChainContention` — so a plain `asyncio.to_thread` from an armed context fails loud, and `wool.to_thread` (a detached fork onto a fresh chain) is the supported alternative.

`ChainContention` carries structured diagnostic fields (`chain_id`, `kind: Literal["thread", "task"]`, plus the offending thread ID or task) and interpolates them into the error message so a stack trace surfaces who collided with whom. `wool.TaskFactoryDisplaced` is raised when a third-party factory installed after Wool's silently drops copy-on-fork for subsequently-created tasks; displacement is detected at three converging points — the factory's `weakref.finalize`, the per-task done-callback backstop, and the `ensure_task_factory_installed` check that every chain mount runs. Because every arming ensures the factory — including the worker's per-frame wire-ingress mount — a worker surfaces a displaced factory at frame ingress, not only on the next caller-side set.

Cooperatively-scheduled work is deliberately not arbitrated. `call_soon` / `call_later` / `call_at`, `add_reader` / `add_writer` / `add_signal_handler`, and `Future.add_done_callback` callbacks inherit the scheduling scope's context via `copy_context()` and run on its chain UUID unchanged. They execute synchronously on the owning thread and never overlap with their scheduler, so sharing the chain is safe.

`install_task_factory` composes with an existing factory via a `__wool_inner__` chain it stamps on its wrapper. The composition-chain walk detects Wool buried under a later third-party install, so `wool → third-party → wool` is recognised and skipped (memoised per layer). A module-level `threading.Lock` guards both the install read-modify-write and the per-task registry against the (narrow) cross-thread install race.

### Frame/ChainManifest wire-frame abstraction

Introduce a `Frame` hierarchy in `wool.runtime.worker.frame` that wraps every request and response on the worker wire:

- `RequestFrame`: `TaskRequestFrame`, `NextRequestFrame`, `SendRequestFrame`, `ThrowRequestFrame`
- `ResponseFrame`: `ResultResponseFrame`, `ExceptionResponseFrame`, `AckResponseFrame`, `NackResponseFrame`

Mid-stream frames mix in a `_ChainManifestFrame` base that captures or accepts a manifest; boundary frames (`TaskRequestFrame`, `AckResponseFrame`, `NackResponseFrame`) carry none. `Frame.chain_manifest` is a union — `ChainManifest | ChainSerializationError | None` — read as "the decoded manifest, the reason it couldn't be parsed, or absent." `Frame.from_protobuf` decodes the optional wire `context` field through `ChainManifest.from_protobuf` and, on a strict-mode decode failure, captures the `ChainSerializationError` as the union's error arm rather than raising at decode time. The failure is deferred to `Frame.mount`, which raises it for a value frame or, when the frame's payload is itself a `BaseException`, walks the payload's `__context__` chain and attaches the decode error at the tail — so a malformed manifest never preempts the routine's primary signal. Each frame owns its own (de)serialisation, giving senders and receivers a single canonical place to encode and decode chain state.

`ChainManifest.from_protobuf` is the pure decoder: it resolves variable identities and deserialises values without ever touching a `contextvars.Context`, emitting `SerializationWarning` per bad entry and aggregating them into one `ChainSerializationError` under strict mode.

Reorganise the protocol package so the codegen has a clean home. Extract the `sys.path` hack, the `wire_pb2` / `wire_pb2_grpc` import block (wrapped as `ProtobufImportError`), and the `AddServicerToServerProtocol` plus `add_to_server` registry into `_wire.py`; the package `__init__` re-exports from it. The wire message is `ChainManifest` (value-only `ContextVar` entries), riding on every `Request` / `Response` as the optional `context` field.

### Per-frame context architecture on the worker

Route every request and response on the worker wire through the new `RequestFrame` and `ResponseFrame` types. Stop threading wire-context application through ad-hoc helpers in `session.py`, `service.py`, and `connection.py`; let each frame carry its own manifest and mount it into the active chain at receive time as a single canonical step. Build outbound traffic in `connection.py` from `RequestFrame` subclasses and consume `ResponseFrame` subclasses on `_read_next`; emit the `service.py` terminal responses (result, exception, ack, nack) through `ResponseFrame`; reframe `session.py`'s request and response paths around the Frame hierarchy. The chain rides only on the mid-stream payload frames (`Next` / `Send` / `Throw` / `Result` / `Exception`), so the first drive frame — not the dispatch metadata — propagates the armed chain.

The dispatch driver caches the worker-side step context per wool chain ID in a `WeakValueDictionary[UUID, contextvars.Context]`. Frames sharing a chain ID reuse one `contextvars.Context`, so a routine's `set` before an async-generator yield and its `reset` after land in the same context — stdlib async-generator parity, since asyncio drives every resumption of one generator in a single context. Frames carrying distinct chain ids resolve to distinct contexts and stay isolated. An unarmed dispatch — one whose caller propagates no chain — runs its frames in a single working context left unarmed, so a stateless routine's plain `asyncio.to_thread` offload copies a bare context and never trips the chain-contention guard; if the routine's own `set` arms that working context, the driver indexes it under the freshly minted chain ID, so the back-propagated next frame (which now carries that ID) reuses it. Because entering a proxy arms the chain, a dispatch routed through a `WorkerProxy` always carries the `_armed` marker, so this unarmed working-context path applies only to dispatches that reach the driver with no propagated chain.

### Routine, discovery, loadbalancer

Update the `@routine` decorator and the underlying `Task` wrapper to read dispatch-time fields off `RuntimeContext` rather than the removed `Context` aliases, and surface `RuntimeContext` on the `Task` itself so call sites can inspect what was actually dispatched. Switch `discovery/local.py` and the matching discovery test to the wire-namespace import (`from wool import protocol`) so they refer to the protobuf `WorkerMetadata` without colliding with the wool dataclass of the same name. Update the loadbalancer exception-handling docstrings to reference the current `ChainSerializationError` shape rather than the removed decode-alias names.

### Build, documentation, and test coverage

Add `uvloop` as a dev dependency and register the `stdlib_parity` pytest marker so the new parity suite runs under both the default asyncio loop and uvloop. Measure coverage from spawned worker processes (`concurrency = multiprocessing,thread`, sigterm flush) so worker-side dispatch code is visible to the gate. Rewrite the context subsystem README, the worker README, and the package README to match the shipped surface — the stdlib-aligned public API, the `Chain.from_manifest` wire-apply pattern, and the Frame-based dispatch vocabulary.

Test coverage spans three families: unit coverage of the new context module in `tests/runtime/context/` (`test_chain.py`, `test_manifest.py`, `test_runtime.py`, `test_var.py`, `test_token.py`, `test_factory.py`, `test_guard.py`, `test_exceptions.py`); a `stdlib_parity` suite in `tests/stdlib_parity/` (`test_context_parity.py`, `test_task_creation.py`, `test_executor_offload.py`, `test_loop_callbacks.py`) pinning Wool's behaviour against the analogous stdlib `contextvars` / `asyncio` surfaces under both loops; and Frame round-trip coverage in `tests/runtime/worker/test_frame.py`, rewritten worker tests across `tests/runtime/worker/`, plus end-to-end coverage in `tests/integration/test_context_var_propagation.py`. The unit suite holds the 98% coverage gate: reachable defensive branches are tested directly, and the genuinely-unreachable guards are marked `# pragma: no cover`.

## Test cases

| 1 | `TestContextVar` | A `ContextVar` with and without a constructor default | `get` is called on an unarmed then armed context | Returns the fallback ladder — supplied default, constructor default, then `LookupError` | `ContextVar.get`, armed-gating |
| 2 | `TestContextVar` | An armed context | `set` then `reset` are called | Value rebinds then restores; the first `set` mints the chain and returns a single-use `Token` | `ContextVar.set` / `reset` |
| 3 | `TestContextVar` | A `Token` from a different chain or var, or an already-used token | `reset` is called with it | Stdlib `contextvars.Token.reset` raises `ValueError` | `ContextVar.reset` rejection via stdlib delegation |
| 4 | `TestContextVar` | Two distinct declarations of one `(namespace, name)` key | The second is constructed | Raises `ContextVarCollision` | Identity model |
| 5 | `TestToken` | A `wool.Token` after a first and a nested `set` | `old_value` is read, then the token is pickled, cloudpickled, and copied | `old_value` reports `MISSING` then the prior value; vanilla `pickle`, `cloudpickle`, and `copy` are rejected with `TypeError` | Token old-value semantics and serialization rejection |
| 6 | `TestToken` | A `wool.Token` (alias of `contextvars.Token`) | Its `repr` is inspected | References the backing `contextvars.ContextVar` | Stdlib Token alias surface |
| 7 | `TestChain` | A `Chain` dataclass with chain ID, vars, reset signals, stub pins | Field access and `mount` field-replacement are exercised | Fields are immutable, equality is identity-based, and `mount` returns an armed copy | Chain dataclass shape |
| 8 | `TestChain` | A chain with bound and reset variables | `to_protobuf` emits the wire manifest | One entry per bound var, reset-var entries, and unserializable values skipped with a `SerializationWarning` (aggregated under strict mode) | Chain wire emission |
| 9 | `TestChainManifest` | A wire `ChainManifest` with values, resets, and an undeclared var | `ChainManifest.from_protobuf` decodes it | Round-trips values and resets, registers a stub for the undeclared var, and raises `ChainSerializationError` on strict-mode failure | Manifest decode |
| 10 | `TestChainMount` | A decoded manifest with a live receiver chain, and an unowned mount under a displaced task factory | `Chain.from_manifest(..., owned=...)` is applied | Drains values into the backings, restamps the thread owner (and the task when owned), unions onto the receiver with incoming winning on overlap, propagates resets, arms the result, and surfaces `TaskFactoryDisplaced` when the factory was displaced | Wire-apply / mount and ingress displacement |
| 11 | `TestRuntimeContext` | A `RuntimeContext` entered with a `dispatch_timeout` override | `__enter__` / `__exit__` and a protobuf round-trip are exercised | The overlay applies on enter and restores on exit; the round-trip preserves an explicit timeout and elides an unset default; double-enter raises | `RuntimeContext` lifecycle and codec |
| 12 | `test_factory.py` | An armed parent task with the factory installed | A child task is created | The child inherits the parent's values on a freshly minted chain UUID | Copy-on-fork |
| 13 | `test_factory.py` | A loop with Wool's factory buried under a later third-party install, or no running loop, or a finalizer firing on an idle loop | `install_task_factory` is re-invoked / called outside a loop / collected | Detects buried Wool via the `__wool_inner__` walk and skips; raises a Wool-flavoured `RuntimeError` outside a loop; treats an idle-loop finalizer as teardown, not displacement | Install entry and idempotency |
| 14 | `test_factory.py` | A Wool factory displaced by a third-party install, by stash, or at task completion | The next `wool.ContextVar` set or task completion runs | Raises `TaskFactoryDisplaced` so the caller knows copy-on-fork is no longer in effect | Displacement detection |
| 15 | `TestChainContention` | An armed chain accessed from a foreign thread | A `wool.ContextVar` `get` / `set` / `reset` runs | Raises `ChainContention` with `kind="thread"` interpolating the chain ID, owner thread, and offender; no false positive when the owner task is pending-off-loop or done | Cross-thread entry guard |
| 16 | `TestChainContentionGuardBoundary` | An armed `contextvars.Context` already driving a live task | A second concurrent task is created on that same context | Raises `ChainContention` with `kind="task"`; a callback on the owning thread does not | No shared context across concurrent tasks |
| 17 | `TestChainContention` (exceptions) | A `ChainContention` instance with structured fields | It is pickled and unpickled | The structured fields and message survive the round-trip; the hierarchy roots under `WoolError` / `WoolWarning` | Exception hierarchy and serialization |
| 18 | `TestWoolToThreadParity` | An armed context | `wool.to_thread` runs a function touching a `wool.ContextVar` | The worker thread runs on a fresh detached chain with no apply-back and no guard trip; a no-op when unarmed | `wool.to_thread` |
| 19 | `TestAsyncioToThreadParity` | An armed context | A plain `asyncio.to_thread` touches a `wool.ContextVar` | Raises `ChainContention` | Cross-thread guard at the stdlib boundary |
| 20 | `TestRunInExecutorParity` | A `wool.ContextVar` set in a scope | `loop.run_in_executor` runs on a thread and a process pool | The executor function carries neither stdlib nor wool context — matching stdlib semantics | Executor offload parity |
| 21 | `TestTaskCreationValuePropagationParity` | A `wool.ContextVar` set in a scope | `asyncio.create_task` / `loop.create_task` / `TaskGroup` create child tasks | The child observes the parent's value on a freshly minted chain UUID (copy-on-fork parity) | Parity matrix (default loop + uvloop) |
| 22 | `TestCallbackValuePropagationParity` | A `wool.ContextVar` set in a scope | Work is scheduled via `call_soon` / `call_later` / `call_at` / `Future.add_done_callback` | Callbacks observe the scheduling scope's value on its shared chain, untouched by the guard | Cooperative scheduling parity |
| 23 | `TestAsyncGeneratorValuePropagation` | An async generator suspended on `yield` in an armed context | The generator advances via `asend` / `athrow` / `aclose` | The generator's chain ID is preserved across resume and the value mutation propagates | Async generator parity |
| 24 | `TestFrameToProtobuf` | A `RequestFrame` or `ResponseFrame` subclass | `for_send` is called with payload and serializer | Returns a `Frame` with payload encoded into the matching protobuf field and the armed chain captured into a `ChainManifest` | Frame construction and emission |
| 25 | `TestFrameFromProtobuf` | A wire envelope, including one whose chain manifest fails to decode under strict mode | `Frame.from_protobuf` is invoked | Returns the matching leaf with payload decoded; a decode failure is captured as the `ChainSerializationError` arm of `chain_manifest`, deferred rather than raised | Frame ingress and deferred decode |
| 26 | `TestChainsDecodeErrorOntoPayload` | An exception, nack, or throw frame carrying a deferred `ChainSerializationError` | `Frame.mount` runs | The decode error chains onto the payload exception's `__context__` tail rather than preempting it | Strict-mode decode chaining |
| 27 | `TestContextVarPropagation` (integration) | A cross-process routine dispatch, including async-generator routines interleaved on one chain | The routine reads and mutates a `wool.ContextVar` | The caller's value propagates in, the worker's mutation propagates back, and serialized interleaving never trips the guard | End-to-end propagation |


wool/src/wool/runtime/context/guard.py
```diff
@@ -0,0 +1,180 @@
...
+# public
+class ChainContention(RuntimeError):
+    """Raised when a Wool chain is entered by a thread or task that does
+    not own it.
+
+    Wool enforces strictly serial execution within a logical chain: at
+    most one OS thread *and* one :class:`asyncio.Task` may run code
+    under a given chain at a time. The guard has two dimensions —
+    :func:`assert_chain_owner` arbitrates OS threads,
+    :func:`assert_owner_task` arbitrates asyncio tasks within a single
```
conradbzura: Rename `assert_chain_owner` to `assert_owning_thread` and `assert_owner_task` to `assert_owning_task`.

wool/src/wool/runtime/context/guard.py
```diff
@@ -0,0 +1,180 @@
...
+    example :func:`asyncio.to_thread`, which copies the surrounding
+    context (chain UUID and all) into the executor thread. The first
+    :class:`wool.ContextVar` access from that thread runs the chain
+    in genuine parallelism with the event loop and raises.
+    :func:`wool.to_thread` is the supported alternative: it forks a
+    fresh, detached chain for the worker thread so the offload never
+    trips the guard.
+
+    The cross-task trigger is two :class:`asyncio.Task` objects
+    sharing one chain. The task factory forks every child task onto a
```
conradbzura: Rephrase "The task factory..." to "Wool's task factory...".

wool/src/wool/runtime/context/snapshot.py
```diff
@@ -0,0 +1,612 @@
...
+    snap_token = install_snapshot(decoded.snapshot)
+    value_tokens = [
+        (var._backing, var._backing.set(value)) for var, value in decoded.values.items()
+    ]
+    try:
+        yield
+    finally:
+        for backing, token in reversed(value_tokens):
+            backing.reset(token)
+        _snapshot.reset(snap_token)
```
conradbzura: Let's make this an instance method on DecodedSnapshot

wool/src/wool/runtime/context/snapshot.py
```diff
@@ -0,0 +1,612 @@
...
+        current: Any = backing.get()
+    except LookupError:
+        # The backing variable must hold *some* value before the dance
+        # reads it: a variable a payload token references but the
+        # snapshot carried no value for starts unset. Seed the
+        # Undefined sentinel — get() treats it as "unset" anyway.
+        backing.set(Undefined)
+        current = Undefined
+    backing.set(token._old_value)
+    token._native = backing.set(current)
```
conradbzura: This should live in `token.py`.

wool/src/wool/runtime/context/snapshot.py
```diff
@@ -0,0 +1,612 @@
...
+        if var_key in encoded_values:
+            entry.value = encoded_values[var_key]
+        if var_key in token_ids_by_key:
+            entry.consumed_tokens.extend(token_ids_by_key[var_key])
+    if failures:
+        raise BaseExceptionGroup(
+            "wool context encode failed for one or more vars",
+            failures,
+        )
+    return wire
```
conradbzura: This should become a `to_protobuf` instance method on `Snapshot`.

wool/src/wool/runtime/context/snapshot.py
```diff
@@ -0,0 +1,612 @@
...
+    # chain.
+    snapshot = Snapshot(
+        chain_id=chain_id,
+        owner=threading.get_ident(),
+        data=data,
+        external_used=external_used,
+        reset_vars=frozenset(reset_vars),
+        stub_pins=frozenset(stub_pins),
+    )
+    return DecodedSnapshot(snapshot=snapshot, values=values)
```
conradbzura: This should also become a `from_protobuf` instance method on `Snapshot`.

wool/src/wool/runtime/context/snapshot.py
conradbzura, line 1: This file is a bit of a mess. I think we can consolidate the two snapshot dataclasses, as well as a number of the helper functions, into a single structured dataclass that addresses our needs.

PR #227, opened 2026-05-19
Remove the passthrough serializer; serialize all dispatch through cloudpickle — Closes #225

### LINKED ISSUE #225: Remove the passthrough serializer; serialize all dispatch through cloudpickle
## Description

Remove `PassthroughSerializer` and its supporting machinery (`_PassthroughKey`, `_passthrough_store`, `_passthrough_pool`) from `wool.runtime.serializer`. Self-dispatch — a worker dispatching a routine that resolves to its own process — currently negotiates `PassthroughSerializer`, which transfers the dispatch payload by object reference instead of serializing it.

After this change, every dispatch — cross-process and self — serializes through `CloudpickleSerializer` (`wool.__serializer__`). The `use_passthrough` branching in `WorkerConnection` (`dispatch`/`_execute`/`_handshake`), the passthrough negotiation in `DispatchSession.__aenter__`, and the `PassthroughSerializer`-specific branches in `Task.to_protobuf`/`from_protobuf` are removed. The dispatch path is left with a single serialization path.

## Motivation

`PassthroughSerializer` is a self-dispatch optimization that hands the routine the *same* argument objects the caller passed, rather than copies. This leaks reference semantics into a routine's contract.

A `@wool.routine` is location-transparent: a cross-process dispatch genuinely copies its arguments through pickle, so a routine reasonably assumes it receives a copy and that mutating an argument in place is local. Under self-dispatch the routine instead shares the caller's objects, so an in-place mutation leaks back to the caller — or does not — depending purely on where the load balancer routed the call. The behavior is observable, non-deterministic, and contradicts `@wool.routine`'s own guidance to avoid shared mutable state across a dispatch.

The most acute symptom is `wool.Token`: it carries a native `contextvars.Token` valid only in the `contextvars.Context` it was minted in. Lent by reference through a same-process dispatch, that native handle is stale on the receiving task and `wool.ContextVar.reset` raises `ValueError`. This is the defect behind the `xfail`'d multi-hop token-reset test on PR #224.

The optimization does not earn its cost. Benchmarked on representative payloads, `cloudpickle` (C-accelerated `pickle` underneath) round-trips ordinary data *faster* than `copy.deepcopy` (2–7×), and serializing a typical self-dispatch payload through cloudpickle costs single-digit microseconds. Passthrough's zero-copy saves only those microseconds, only for self-dispatch — at the price of a correctness leak, a real bug, and a permanent special-case branch threaded through the connection, session, and task layers. Removing it is a net simplification.

## Expected outcome

- `PassthroughSerializer`, `_PassthroughKey`, `_passthrough_store`, and `_passthrough_pool` are removed from `wool.runtime.serializer`.
- All dispatch — cross-process and self-dispatch — serializes through `CloudpickleSerializer`; a routine always receives a copy of its arguments, regardless of where it runs.
- The `use_passthrough` branches in `WorkerConnection` and `DispatchSession`, and the `PassthroughSerializer` handling in `Task.to_protobuf`/`from_protobuf`, are removed; one serialization path remains.
- Self-dispatch behaves observationally identically to cross-process dispatch — including that an argument which is not cloudpicklable can no longer be carried (cross-process dispatch never could).
- The multi-hop `wool.Token` reset scenario is correct once this lands; PR #224 removes its `TestMultiHopTokenReset` `xfail` when it takes this change in.
- The test suite stays green; tests cover the unified self-dispatch path.

conradbzura:
## Summary

`PassthroughSerializer` was a self-dispatch optimization: when a worker dispatched a routine that resolved to its own process, the payload (callable, args, kwargs, proxy) was stashed by reference in a module-level `WeakKeyDictionary` and retrieved with a UUID token rather than serialized. This leaked reference semantics into a routine's contract — a `@wool.routine` is location-transparent, so a routine reasonably assumes it receives a copy of its arguments, but under self-dispatch it instead shared the caller's objects. An in-place mutation of an argument therefore leaked back to the caller, or did not, depending purely on where the load balancer routed the call. The acute symptom was `wool.Token`, whose native `contextvars.Token` is valid only in its minting context: lent by reference through a same-process dispatch it was stale on the receiving task and `wool.ContextVar.reset` raised `ValueError`.

Remove `PassthroughSerializer` and its supporting machinery so every dispatch — cross-process and self — serializes through `CloudpickleSerializer` (`wool.__serializer__`). A routine now always receives a copy of its arguments, regardless of where it runs. Because the wire serializer-negotiation machinery existed solely to carry a non-default serializer for passthrough, remove it too: the `serializer` parameter of `Task.to_protobuf`/`from_protobuf`, the `_pickle_serializer`/`_unpickle_serializer` helpers, and the `Task.serializer` protobuf field. The pluggable `Serializer` protocol and `CloudpickleSerializer` remain as the process-wide default.

The trade-off is the loss of a zero-copy micro-optimization on the self-dispatch path, which costs single-digit microseconds per self-dispatch — cheap relative to the correctness leak, the real `wool.Token` defect, and the permanent special-case branch threaded through the serializer, connection, session, and task layers that it removes.

Closes #225

### Remove `PassthroughSerializer` and its supporting machinery

Delete `PassthroughSerializer`, `_PassthroughKey`, the `_passthrough_store` weak-key map, and the `_passthrough_pool` resource pool from `wool.runtime.serializer`. `CloudpickleSerializer` and the `Serializer` protocol stay.

### Unify `Task` serialization

`Task.to_protobuf` loses its `serializer` parameter and unconditionally serializes the callable, arguments, and proxy through `wool.__serializer__`. `Task.from_protobuf` loses its serializer-field branch and unconditionally deserializes through `cloudpickle.loads`. Delete the `_pickle_serializer`/`_unpickle_serializer` LRU-cached helpers.

### Remove the `use_passthrough` branching from `WorkerConnection`

`dispatch`, `_execute`, and `_handshake` no longer negotiate or pin a passthrough serializer. Self-dispatch detection and the Unix-domain-socket transport routing are unchanged — only the serializer choice is unified.

### Remove passthrough negotiation from `DispatchSession`

`DispatchSession.__aenter__` no longer sniffs the wire `serializer` field or acquires a per-task serializer from a pool; `DispatchSession.serializer` is always `wool.__serializer__`.

### Drop the `Task.serializer` wire field

Remove field 12 from the `Task` protobuf message (reserve the number) and regenerate the bindings. **Breaking change:** the `Task.serializer` wire field no longer exists and `Task.to_protobuf` no longer accepts a `serializer` argument.

### Add coverage for the unified dispatch path

Add unit and integration tests asserting the new guarantees: a dispatch yields a copy of its arguments, a non-cloudpicklable argument fails self-dispatch the same way it has always failed cross-process, and `wool.Token` reset works under self-dispatch.

## Test cases

| 1 | `TestTask` | A Task with picklable callable, args, kwargs, and proxy | The task round-trips through `to_protobuf` and `from_protobuf` | Each restored field is a distinct object from the original | Copy semantics replace passthrough identity |
| 2 | `TestTask` | A Task whose args hold a mutable object | The restored arg is mutated after a round-trip | The original arg is unaffected | Round-trip copy independence |
| 3 | `TestTask` | A Task whose args hold a non-cloudpicklable value | `to_protobuf` is called | A pickling error is raised | Unpicklable positional argument |
| 4 | `TestTask` | A Task whose kwargs hold a non-cloudpicklable value | `to_protobuf` is called | A pickling error is raised | Unpicklable keyword argument |
| 5 | `TestTask` | A fully populated Task | `to_protobuf` is called | The message exposes no `serializer` field | Reserved protobuf field 12 |
| 6 | `TestTask` | Any nested cloudpicklable structure as args and kwargs | The task round-trips through protobuf | The restored payload equals the original yet is non-identical | Copy invariant (property-based) |
| 7 | `TestTask` | A protobuf Task whose callable field holds non-cloudpickle bytes | `from_protobuf` is called | An unpickling error is raised | Corrupt payload rejection |
| 8 | `TestCloudpickleSerializer` | A serializer class overriding `__eq__` without `__hash__` | `isinstance` against the Serializer protocol is evaluated | It returns False | Serializer protocol conformance |
| 9 | `TestCloudpickleSerializer` | A CloudpickleSerializer and a module-level callable | The callable is dumped then loaded | It restores to a callable producing the same result | Callable round-trip |
| 10 | `TestWorkerConnection` | A self-dispatch connection with a ContextVar set | `dispatch` sends the initial request | The request's context vars are cloudpickle-encoded and round-trip | Self-dispatch var serialization |
| 11 | `TestWorkerConnection` | A cross-process connection | `dispatch` sends the initial request | The task payload cloudpickle-decodes to a coroutine function | Cross-process payload serialization |
| 12 | `TestWorkerConnection` | Any picklable ContextVar value on the caller context | `dispatch` writes the request | The decoded var value equals the original | Var round-trip (property-based) |
| 13 | `TestDispatchSession` | A first frame carrying an async-generator Task | The session is entered | `serializer` is `wool.__serializer__` and `streaming` is True | Cloudpickle parse of a streaming task |
| 14 | `TestDispatchSession` | A session whose `drain` raises CancelledError | The `async with` block exits | CancelledError propagates and the exit stack still unwinds | Drain registered as an exit-stack callback |
| 15 | `TestDispatchSession` | A session cancelled mid-teardown | Caller cancellation arrives during `__aexit__` | `drain` runs to completion before CancelledError propagates | Shielded teardown unwind |
| 16 | `TestUnifiedDriverShape` | A routine mutating a list argument in place, DEFAULT pool | The caller dispatches passing a referenced list | The caller's list is unchanged | Self-dispatch argument copy |
| 17 | `TestUnifiedDriverShape` | The same routine, EPHEMERAL pool | The caller dispatches passing a referenced list | The caller's list is unchanged | Cross-process argument-copy baseline |
| 18 | `TestUnifiedDriverShape` | An unpicklable argument, DEFAULT pool | The caller dispatches the routine | Dispatch raises at the serialization boundary | Self-dispatch unpicklable rejection |
| 19 | `TestUnifiedDriverShape` | An unpicklable argument, EPHEMERAL pool | The caller dispatches the routine | Dispatch raises at the serialization boundary | Cross-process unpicklable baseline |
| 20 | `TestUnifiedDriverShape` | An async generator mutating an argument across yields | The caller iterates to exhaustion | The caller's argument is unchanged | Streaming argument copy |
| 21 | `TestSelfDispatchTokenReset` | A caller-minted `wool.Token`, DEFAULT pool | The routine resets the token on the worker | The reset does not raise ValueError | Self-dispatch token reset |
| 22 | `TestSelfDispatchTokenReset` | A caller-minted token reset under DEFAULT and EPHEMERAL pools | The dispatch runs under each pool mode | Both produce identical results | Self vs cross-process parity |
| 23 | `TestSelfDispatchTokenReset` | A worker-minted `wool.Token`, DEFAULT pool | The caller resets the token locally | The first reset succeeds and a second raises RuntimeError | Worker-minted token round-trip |
| 24 | `TestSelfDispatchTokenReset` | A caller-minted token consumed in an async-generator frame | The generator is iterated to completion | The mid-stream reset does not raise and reuse raises RuntimeError | Streaming token reset |


wool/proto/wire.proto
```diff
@@ -18,7 +18,7 @@ message Task {
     bytes kwargs = 9;
     int32 timeout = 10;
     optional RuntimeContext runtime_context = 11;
-    optional bytes serializer = 12;
+    reserved 12;  // was: optional bytes serializer (removed in #225)
```
conradbzura: Just drop this — we aren't maintaining backwards compatibility here anyway.

wool/tests/integration/test_context_var_propagation.py
```diff
@@ -876,6 +876,163 @@ async def body():
...
+    Before issue #225 removed the passthrough serializer, self-dispatch
+    handed the worker the caller's same argument objects by reference.
+    A :class:`wool.Token` carries a UUID scoped to the
+    :class:`wool.Context` it was minted in; passed by reference into a
+    self-dispatched routine running in a different Context, the token's
+    handle went stale and ``ContextVar.reset`` raised
+    :class:`ValueError`. With #225, self-dispatch serializes the
+    payload through cloudpickle exactly like cross-process dispatch, so
+    a routine is location-transparent and token reset behaves
+    identically regardless of pool mode.
```
conradbzura: Drop the history lesson and focus on what the suite covers NOW.

wool/tests/integration/test_unified_driver.py
```diff
@@ -800,3 +796,209 @@ async def consume():
...
+        self, credentials_map, retry_grpc_internal
+    ):
+        """Test self-dispatch hands the routine a copy of a mutable argument.
+
+        Given:
+            An in-place-mutating coroutine routine and a DEFAULT pool
+            (single in-process worker so the dispatch target matches
+            the caller process; since issue #225 removed the
+            passthrough serializer, self-dispatch serializes the
+            payload through cloudpickle).
```
conradbzura: Drop the history lesson (i.e., reference to #225).

Issue #336, opened 2026-07-27
Stop resource_tracker KeyError spam when attaching shared memory on Python below 3.13


conradbzura:
## Description

On Python versions below 3.13, `wool` prints bare `resource_tracker` tracebacks to stderr during normal worker-pool operation. Nothing fails — exit codes stay 0 and all work completes — but dozens of tracebacks scroll past, which reads as a crash.

```
Traceback (most recent call last):
  File ".../python3.12/multiprocessing/resource_tracker.py", line 264, in main
    cache[rtype].remove(name)
KeyError: '/iV4C1iXslIkIXmhWzi8gAQdA_fwYNR'
```

This is a regression in `0.12.0rc10`, introduced by 49c4f2e ("fix: Refresh existing registration when a worker is re-added"). It does not occur in `0.12.0rc9`.

Reproduction:

```sh
uv venv v && VIRTUAL_ENV=$PWD/v uv pip install "wool==0.12.0rc10" torch
./v/bin/python .blog/posts/hello_world.py 2>&1 | grep -c resource_tracker   # 77
```

Controlled A/B on the same machine, same virtualenv, same script, CPython 3.12.10, swapping only the wool version:

| Version | Occurrences | Kind |
|---|---|---|
| `0.12.0rc9` | 4 | `UserWarning` — leaked-object warnings at shutdown, step 9 only |
| `0.12.0rc10` | 77 | `KeyError` tracebacks, every step |

These are distinct phenomena, not more of the same: the rc9 warnings come from CPython's own shutdown sweep, while rc10 emits raw tracebacks. Counts scale with pool size at roughly two per worker — across the nine steps: 2, 2, 2, 2, 15, 18, 16, 20.

Confirming the source diverged: `grep -c resource_tracker wool/src/wool/runtime/discovery/local.py` returns 2 at `cf5842f` (rc10) and 0 at `cc6f5ab` (rc9).

Two factors make this worse than ordinary noise. The affected branch carries `# pragma: no cover`, so it is untested — yet it is the branch every supported interpreter below 3.13 takes, and the package requires `>=3.11`. And because this is stderr output rather than a failure, no exit-code-based CI check can catch it.

## Expected behavior

Attaching to an existing shared-memory segment on Python below 3.13 produces no `resource_tracker` output, however many times the same segment is attached and detached within a process.

[...]

Issue #316, opened 2026-07-16
Bound LocalDiscovery's lock acquisition and stop its busy-spin


conradbzura:
## Description

`LocalDiscovery`'s cross-process file lock has two defects that compound each other: it busy-spins while waiting, and it waits forever.

The docstring of `_lock` (`src/wool/runtime/discovery/local.py:879`) states that it "Retries every 1ms until the lock is acquired." It does not. The retry does `await asyncio.sleep(0)`, which yields to the event loop and immediately reschedules — a hot spin, not a 1ms poll. Every iteration issues a real `fcntl` syscall through `portalocker`, so a contended lock pins a core for the duration of the wait. The comment is the only thing that would tell a reader otherwise, and it has been wrong since `c3c197b` ("feat: Add local shared-memory discovery").

Spinning cannot help here. The lock exists to serialize *unrelated processes* against one shared-memory segment, so the holder is by definition in another process. Yielding fast in this process's event loop does nothing to make the holder release sooner — it only burns CPU competing with the very process being waited on.

Separately, the acquisition loop is `while True:` with no timeout, so a holder that never releases wedges the caller indefinitely. This is reachable: `WorkerPool` teardown with `shutdown_timeout=None` and a wedged publish never returns, and the worker is never stopped (verified — `stop` awaited 0 times, `__aexit__` never returns). With a finite `shutdown_timeout`, #298's fix already converts this into a logged failure and a reaped worker, so bounding the lock closes the remaining hole rather than duplicating that work.

`_lock` has exactly one call site — `Publisher.publish` (`local.py:441`) — so the blast radius is small. `publish` is a cold path, called once per worker for `worker-added` and once for `worker-dropped`, never per dispatch, so a 1ms poll costs nothing that is measurable against process spawn or a 60s shutdown budget.

## Expected behavior

Retry at the 1ms interval the docstring already promises, so the fix makes the existing documentation true rather than requiring a doc change.

Bound acquisition with a timeout exposed as a `LocalDiscovery` constructor kwarg — `lock_timeout: float | None = 30.0`, where `None` preserves today's wait-forever behavior — plumbed through to the nested `Publisher` exactly as `block_size` already is (`local.py:239` → `:325` → `:396`).

Expiry raises `TimeoutError`, which composes with pool teardown as-is: `publish` raises, teardown logs it against the worker uid, and the worker is still reaped.

[...]

PR #341, opened 2026-07-27
Stop a failed shared-memory create from emitting a stray resource_tracker unregister — Closes #106, #340

### LINKED ISSUE #106: Add WorkerPool constructor overload that accepts DiscoverySubscriberLike to support external-only discovery patterns
## Description

Extend `WorkerPool` to accept `DiscoverySubscriberLike` in addition to `DiscoveryLike` for the `discovery` parameter. When only a subscriber is provided (no publisher), the pool operates in external mode only — it discovers externally managed workers but cannot spawn or publish its own. If `spawn > 0` is requested with a subscriber-only discovery, the pool rejects it.

```python
# Full discovery — supports ephemeral, external, and hybrid modes (unchanged)
async with WorkerPool(spawn=4, discovery=SomeFullDiscovery()):
    ...

# Subscriber-only — external mode only
async with WorkerPool(discovery=some_discovery.subscriber):
    ...

# Subscriber-only + spawn > 0 — rejected by type checkers at the call site, TypeError at context entry
async with WorkerPool(spawn=4, discovery=some_discovery.subscriber):
    ...
```

## Motivation

Some discovery backends are inherently observe-only. For example, a Docker discovery subscriber watching Swarm-managed containers can detect workers as they appear and disappear, but has no meaningful publish step — the Swarm scheduler owns container lifecycle and labeling. Forcing these backends to implement a no-op publisher to satisfy the `DiscoveryLike` protocol is unnecessary ceremony. Accepting a bare `DiscoverySubscriberLike` makes external-only backends a first-class concept and lets the pool reject an incompatible backend with a type error rather than failing silently at runtime.

This change is fully backwards compatible — existing code passes `DiscoveryLike` instances which satisfy both paths.

## Expected outcome

- The `discovery` parameter on `WorkerPool` accepts `DiscoveryLike | DiscoverySubscriberLike`, or any `Factory` form resolving to one, for a pool with no `spawn`.
- When a `DiscoverySubscriberLike` without a publisher is provided, the pool operates in external mode: it subscribes to discover existing workers but does not publish.
- When `spawn > 0` is combined with a subscriber-only discovery, static type checkers reject the call, and the pool raises `TypeError` at context entry, where `Factory` forms first resolve to an object.
- No changes to existing `DiscoveryLike` implementations or callers.

### LINKED ISSUE #340: Stop a failed shared-memory create from emitting a stray resource_tracker unregister
> **Correction (2026-07-28).** The version scope below was wrong. The fault reproduces on **every supported version**, 3.13 included: `track` defaults to `True` for a create, so 3.13's `self._track` guard on `SharedMemory.unlink` is satisfied and the stray unregister is emitted anyway. Verified by probe on 3.11.12, 3.12.10 and 3.13.3. The text is corrected in place; see #341 for the fix, which therefore carries no version gate.

## Description

A `SharedMemory` construction that fails partway through emits an `unregister` for a name the resource tracker never held. That produces the same `KeyError` traceback #336 removed — printed by the tracker process, invisible to the exit code.

Reproduced on CPython 3.11.12, 3.12.10 and 3.13.3 alike, by forcing the truncation or the mapping to fail:

```
create raised: OSError
unregisters emitted by the failed create: [('shared_memory', '/wool-probe-create')]
```

```
Traceback (most recent call last):
  File ".../python3.12/multiprocessing/resource_tracker.py", line 264, in main
    cache[rtype].remove(name)
KeyError: '/wool-probe-create'
```

Two call sites in `wool/src/wool/runtime/discovery/local.py` construct with `create=True` and are exposed: `LocalDiscovery.__enter__` and `_shared_memory_factory`. The attach path is already guarded — `_attach` rebinds both tracker hooks for the duration of its constructor call, as of #336 — but the create paths do not go through it.

Reaching this requires a failure inside the constructor's guarded block: `ENOSPC` from the truncation on an exhausted `/dev/shm`, or `ENOMEM` from the mapping. A `shm_open` failure raises ahead of that block and is unaffected. It is rare, pre-existing, and unrelated to the re-add traffic that made #336 loud.

There is a second half that wool cannot fix. On an *attach*, the same handler also calls `shm_unlink`, destroying a segment this process does not own while its creator is still running. `_attach` suppresses the tracker half but cannot intercept the unlink — that behaviour belongs to CPython ([bpo-38119](https://github.com/python/cpython/issues/82300)) and is noted here only so the residual is recorded rather than rediscovered.

## Expected behavior

A `SharedMemory` construction that fails partway through leaves the resource tracker's cache exactly as it found it, whichever call site made it. No `KeyError` reaches the tracker process, and no traceback reaches stderr.

This holds on every supported version. Unlike the attach case, there is no version on which the standard library already does the right thing: `track=False` prevents the unregister for an attach, but a create must stay tracked, so the guard cannot be expressed through that parameter.

## Root cause

`SharedMemory.__init__`, CPython 3.12:

```python
try:
    if create and size:
        os.ftruncate(self._fd, size)
    stats = os.fstat(self._fd)
    self._mmap = mmap.mmap(self._fd, size)
except OSError:
    self.unlink()
    raise

resource_tracker.register(self._name, "shared_memory")
```

`register` runs *after* the `try`, so a failure in `ftruncate`, `fstat`, or `mmap` reaches `unlink()` with the name never registered.

Below 3.13 `unlink` is unconditional:

```python
def unlink(self):
    if _USE_POSIX and self._name:
        _posixshmem.shm_unlink(self._name)
        resource_tracker.unregister(self._name, "shared_memory")
```

On 3.13 and up it is guarded — but by a flag a create sets:

```python
def unlink(self):
    if _USE_POSIX and self._name:
        _posixshmem.shm_unlink(self._name)
        if self._track:
            resource_tracker.unregister(self._name, "shared_memory")
```

`self._track` is `track`, which defaults to `True`. A create takes that default, so the guard passes and the stray unregister is emitted on 3.13 exactly as on 3.12.

The tracker's cache is a per-type set; removing an absent name raises `KeyError` in its main loop, which prints a traceback and sets an exit code nothing waits on.

The fix extends #336's shim to wool's own `create=True` sites — narrower than the attach case, since there the name genuinely should end up registered once construction succeeds.

**Update: the remedy has changed.** That shim was implemented in #341 and reviewed twice; the review found the workaround had grown more defects than the fault it guards against. The direction is now #353 — back the registry with a memory-mapped file and remove the resource tracker from the picture entirely, which makes this fault unreachable by construction rather than suppressed. The analysis above stands; only the remedy is superseded.

conradbzura:
## Summary

Route Wool's two `SharedMemory(create=True)` sites through a helper that suppresses the stray `resource_tracker.unregister` a construction emits when it fails partway through.

`SharedMemory.__init__` unlinks from inside the `except OSError` handler wrapping the block that truncates, stats and maps the segment, and `register` runs *after* that block. Below 3.13 `unlink` unregisters unconditionally; on 3.13 and up it unregisters whenever `track` is set, which it is for a create. So a construction that fails mid-way unregisters a name that was never registered. The tracker's per-type cache is a set, so removing an absent name raises `KeyError` in the tracker process, which prints a traceback to stderr and sets an exit code nothing waits on — the same observable #336 removed from the attach path, reached instead by the truncation or the mapping failing.

**This corrects the issue's stated scope**, and #340 has been corrected in place. The issue said 3.13 was unaffected because `unlink` is guarded by `self._track`. That guard does not help a create: `track` defaults to `True`, so it is satisfied. Probed directly on all three supported interpreters:

| | 3.11.12 | 3.12.10 | 3.13.3 |
|---|---|---|---|
| failed create emits a stray unregister | yes | yes | **yes** |
| tracker `KeyError` traceback on stderr | yes | yes | yes |

The fix therefore carries **no version gate**, unlike #336's attach fix. That is also its best CI property: the new code runs on every matrix leg with no fixture forcing a version the interpreter is not on.

The `shm_unlink` half that #336 had to leave to CPython (bpo-38119) has no counterpart here. On an attach it destroys a segment the process does not own; on a create it destroys the one the call just made, which is correct. Nothing is left over.

Closes #106 
Closes #340

### Extract the suppression window from `_attach`

`_attach` rebound both tracker hooks for one constructor call. A create needs the same window but a different policy — the segment must stay *registered* on success, since that entry is what its own `unlink` removes; only the failure path is wrong.

Move the machinery into a `_suppressing(name, *, register, unregister=True)` context manager. Name the parameters for the hooks rather than switching on a single boolean, so both call sites read without consulting the callee. Leave `_attach` a two-line caller that keeps its version gate. Rename `_attach_lock` and `_reinit_attach_lock` to `_tracker_lock` and `_reinit_tracker_state`, since they no longer serve attaches alone. Committed separately as a behaviour-preserving refactor.

### Harden the window against three hazards review surfaced

- **Restore what this module displaced, not an import-time snapshot.** The at-fork handler previously reset the hooks unconditionally, so *any* fork in a process that had imported Wool clobbered instrumentation another library installed after import — even with no window ever opened. Track each shim with the hook it displaced and restore only that.
- **Neutralise a shim before unwinding.** A third party that wraps the hook inside the window keeps our shim in its delegation chain for the life of the process, where it would go on matching this segment's name against a thread ident the interpreter is free to reuse. A liveness flag makes a stranded shim a pass-through.
- **Bound the acquisition.** `DEFAULT_LOCK_TIMEOUT` already governs this module's other lock, and #316 bounded that one for the same reason. An unbounded acquire turns a wedged holder into a process-wide stall with nothing to diagnose.

### Add `_create` and route both call sites

`_create(name, size)` opens the window over `unregister` only. `LocalDiscovery.__enter__` and `LocalDiscovery.Publisher._shared_memory_factory` both use it.

`__enter__`'s `FileExistsError` fallback to `_attach` is unaffected: the exception propagates out of the context manager, releasing the non-reentrant lock, before the handler runs. Pin that with a subprocess test rather than leaving it to inspection — a refactor that widened the window across the whole `try`/`except` deadlocks rather than fails, and three of the fifteen reviewers independently wrote exactly that refactor while reviewing.

### Give the existing attach assertion teeth

`test__attach_should_suppress_the_unregister_when_the_mapping_fails` asserted only that the ledger recorded no violation — and **passed with the suppression removed entirely**. The creator's entry is live at that point, so an unsuppressed unregister is absorbed as an ordinary discard and records nothing. Assert on the residual set instead, where the creator's registration either survives or does not.

### Test design notes

**Inject at `os.ftruncate`, not `mmap.mmap`.** Only a create truncates, so failing it breaks creates and leaves every attach working. That matters because `publish` remaps the address space *before* reaching the block factory; a broken mapping would take out the arrangement rather than the target.

**Seed the name before every failure test.** The `tracker_ledger` fixture classifies an unregister only for a name it has seen registered. A first-time failed create names something it never saw, so the stray is forwarded unclassified — leaving the assertion green whether or not the fix is present, while planting a real `KeyError` in the session's tracker.

**Run the liveness properties in a subprocess.** The lock is process-global, so a wedge inside pytest strands every later test in the session; a wedge in a child is reaped by the harness timeout and reported against the test that caused it.

**Derive the page boundary rather than hard-coding it.** The capacity property crosses the point where a segment outgrows one page, which is where an under-sized request first becomes observable. That point is capacity 255 on Linux's 4 KiB pages and 1023 on macOS arm64's 16 KiB pages, so a fixed example tuned to either is a no-op on the other.

**Carry no version dimension on the create tests.** The fault reproduces unpatched on every supported version, so a legacy/native split would run identical code twice and encode a conditionality that does not exist.

### Mutation results

Each mutant, and the tests that caught it:

| Mutant | Caught by |
|---|---|
| Create suppression removed | 2 unit + 2 integration |
| `register` suppressed on the create path | 7 unit |
| Unregister shim forwards unconditionally | 6 unit, incl. the strengthened #336 test |
| `__enter__` site left unrouted | 1 unit + 1 integration |
| Block site left unrouted | 1 unit + 1 integration |
| Name clause dropped from the match | 4 unit |
| Resource-type clause dropped from the match | 3 unit |
| Restore only on the returning path | whole suite wedges — the lock is never released |
| Restore only one of the two hooks | 1 unit |
| Identity guard dropped from the restore | 1 unit |
| Segment created one byte long | 1 unit |
| `__enter__` halves the size it requests | 1 unit |
| Suppression window widened across the fallback | 1 integration |
| At-fork registration deleted | 1 integration |
| At-fork handler stops replacing the lock | 1 integration |

The last eight all survived the first round of tests; review found them, and they now fail. Both unrouted-site mutants are caught by exactly one unit test and one integration test each, so the two call sites have independent guards.

### Coverage

Unit 98.86% against a floor of 98; integration 86.46% against 70. `local.py` holds at 6 missed lines — the same pre-existing set as before this change, with the `_unregister` pass-through gap inherited from #336 now covered and the new bounded-acquire path covered by a test rather than a pragma. No `pragma` added or removed.

Subprocess children earn no coverage credit, so the integration tests are behavioural evidence and the unit suite carries the coverage. That division is stated in the integration module docstring.

## Test cases

| 1 | `TestLocalDiscovery` | A namespace no process currently owns | An owner enters its context and leaves it | It should hold exactly one tracked segment for the life of the context and none after it | Create success path; registration deliberately not suppressed |
| 2 | `TestLocalDiscovery` | A namespace an owner has already entered and left, and a filesystem that fails the truncation | A fresh owner enters that namespace | It should propagate the error without unregistering a name the tracker is not holding | Create failure path at `LocalDiscovery.__enter__` |
| 3 | `TestLocalDiscovery` | A namespace whose create already failed, and a filesystem that has since recovered | That same namespace is entered again | It should track the retried segment — the same name is what a stranded shim would still match | Hook restoration, scoped to the failed name |
| 4 | `TestLocalDiscovery` | An owner whose create already failed | A second, unrelated namespace is entered and left | It should track that segment and release it on exit | Restoration is not name-wide |
| 5 | `TestLocalDiscovery` | Any capacity spanning the point at which the address space outgrows one page | The namespace is entered and every declared slot is written and read back | It should map the whole address space, whatever the capacity | `_create`'s `size` argument |
| 6 | `TestLocalDiscovery` | A construction that unregisters an unrelated segment while the window is open | An owner enters its context | It should let that unregister reach the tracker | Suppression narrowed to one segment |
| 7 | `TestLocalDiscoveryPublisher` | An owned namespace and a publisher on it | A worker is added and then dropped | It should track the block alongside the address space and release it on the drop | Create success path at the pool factory |
| 8 | `TestLocalDiscoveryPublisher` | An owned namespace, a worker already added and dropped, and a filesystem that fails the truncation | That worker is added again | It should propagate the error without unregistering a name the tracker is not holding | Create failure path at the pool factory |
| 9 | `TestLocalDiscoveryPublisher` | An arbitrary sequence of add, update and drop events, and an arbitrary subset of them whose block creation fails | The sequence is published serially and the owner exits | It should never unregister a name the tracker is not holding, whichever creates failed | The balance invariant, now including failures |
| 10 | Module-level `_suppressing` | An open window, and any resource differing in name, in type, or in both | That resource is registered and unregistered inside the window | It should let both calls through | Narrowing on name and resource type, both configurations |
| 11 | Module-level `_suppressing` | An open window whose block raises | The exception propagates | It should leave a later segment tracked and hold no shim | Restoration on the raising path |
| 12 | Module-level `_suppressing` | An open window whose block rebinds the tracker's hook | The window closes | It should leave that wrapper installed | The identity guard's documented promise |
| 13 | Module-level `_suppressing` | Another thread holding the lock, and a short bound | A second window is opened | It should raise `TimeoutError` naming the wait | The bounded acquisition |
| 14 | Module-level `_create` | An attach of a name no segment holds, followed by a create of that same name | The attach raises and the create follows | It should track the created segment | Cross-path interaction on one name |
| 15 | `TestCrossProcessTracker` | An independent interpreter whose filesystem fails the truncation | It fails to create the namespace segment and exits | It should exit 0 with no tracker traceback on stderr | `__enter__`, across a real process boundary |
| 16 | `TestCrossProcessTracker` | The same, failing one worker announcement | The publish raises and the interpreter exits | It should exit 0 with no tracker traceback on stderr | The pool factory, across a real process boundary |
| 17 | `TestCrossProcessTracker` | A namespace this process owns, and an interpreter entering it as a non-owner | Its create raises `FileExistsError` and the attach runs | It should attach and exit within the bound | The lock is released before the fallback |
| 18 | `TestCrossProcessTracker` | An interpreter with a thread parked inside an open window | It forks and the child enters a namespace of its own | It should report the child entering and exiting 0 | The at-fork handler |
| 19 | `TestCrossProcessTracker` | An interpreter creating a segment larger than the system can back, patching nothing | It runs to completion and its tracker drains | It should exit 0 yet print a tracker `KeyError` | Negative control, independent of both Wool and the injection technique |
| 20 | Module-level `_attach` | A segment created by this process and a mapping that raises after it is opened | The segment is attached by name | It should leave the creator's registration in the tracker's cache | Strengthened #336 assertion, now shared with the create path |

### DIFF
--- a/wool/src/wool/runtime/discovery/local.py
+++ b/wool/src/wool/runtime/discovery/local.py
@@ -229,9 +229,12 @@ class LocalDiscovery(Discovery):
 
     **Lifecycle.** An instance is single-use: it may be entered once, and a
     second entry raises `RuntimeError` whether or not the first has exited.
-    Use a fresh instance per ``with`` block. The guard binds to the instance,
-    so distinct instances sharing a namespace — which compare equal — are
-    unaffected by each other.
+    An entry that *failed* also counts — a create that raises `OSError`
+    leaves nothing behind, but the instance is spent, so recovering from a
+    transient failure means retrying with a fresh instance rather than the
+    same one. Use a fresh instance per ``with`` block. The guard binds to
+    the instance, so distinct instances sharing a namespace — which compare
+    equal — are unaffected by each other.
 
     The owner's exit unlinks the segment out from under every still-attached
     peer, while a non-owner's exit only closes its own mapping; a non-owner
@@ -323,15 +326,18 @@ class LocalDiscovery(Discovery):
         :returns:
             This instance.
         :raises RuntimeError:
-            If this instance has already been entered.
+            If this instance has already been entered — including by an
+            entry that raised, which still consumes the instance.
+        :raises OSError:
+            If the namespace's segment can neither be created nor
+            attached. Transient at the OS level, but recovery needs a
+            fresh instance; see `LocalDiscovery` for why.
+        :raises TimeoutError:
+            If the tracker rebind window cannot be entered; see `_create`.
         """
         size = _HEADER_SIZE + self._capacity * REF_WIDTH
         try:
-            self._address_space = SharedMemory(
-                name=_short_hash(self._namespace),
-                create=True,
-                size=size,
-            )
+            self._address_space = _create(_short_hash(self._namespace), size)
             self._owner = True
         except FileExistsError:
             self._address_space = _attach(_short_hash(self._namespace))
@@ -547,7 +553,14 @@ class LocalDiscovery(Discovery):
                 registration is restored before the error propagates.
             :raises TimeoutError:
                 If the cross-process file lock is not acquired within this
-                publisher's ``lock_timeout``.
+                publisher's ``lock_timeout``, or the tracker rebind window
+                within `DEFAULT_LOCK_TIMEOUT`; see `_create` for the
+                distinction.
+            :raises OSError:
+                For ``worker-added``, if the worker's block cannot be
+                created — e.g. an exhausted ``/dev/shm``. Nothing is
+                registered and the publisher stays usable, so a retry may
+                succeed.
             """
             async with _lock(self._namespace, timeout=self._lock_timeout):
                 with _shared_memory(_short_hash(self._namespace)) as address_space:
@@ -686,13 +699,12 @@ class LocalDiscovery(Discovery):
                 The name for the shared memory block (typically a worker UUID
                 hex string).
             :returns:
-                A new SharedMemory instance.
+                A new SharedMemory instance, tracked by this process.
+            :raises OSError:
+                If the segment cannot be created. The creation goes through
+                `_create`; see it for the tracker and failure semantics.
             """
-            shared_memory = SharedMemory(
-                name=name,
-                create=True,
-                size=self._block_size,
-            )
+            shared_memory = _create(name, self._block_size)
 
             def cleanup():  # pragma: no cover
                 _unlink_quietly(shared_memory)
@@ -1146,6 +1158,63 @@ def _suppressing(
         lock.release()
 
 
+def _create(name: str, size: int) -> SharedMemory:
+    """Create a shared memory segment, tracked by this process.
+
+    This process owns the segment, and the registration this makes is what
+    its own later `unlink` removes. Only the tracker call a *failed*
+    construction would make is silenced.
+
+    :param name:
+        The name of the shared memory segment to create.
+    :param size:
+        The size of the segment, in bytes.
+    :returns:
+        A tracked `SharedMemory` mapped to a newly created segment.
+    :raises FileExistsError:
+        If a segment of that name already exists. Raised before anything
+        is mapped, so nothing is left behind; `LocalDiscovery.__enter__`
+        routes this case to `_attach`.
+    :raises OSError:
+        If the truncation, stat or mapping fails partway through
+        construction — ``ENOSPC`` on an exhausted ``/dev/shm``, ``ENOMEM``
+        on the mapping. The segment is destroyed by the constructor's own
+        handler before the error escapes, so the caller has nothing to
+        clean up; this is the failure the suppression exists for.
+    :raises TimeoutError:
+        If the rebind window cannot be entered within
+        `DEFAULT_LOCK_TIMEOUT`. This bound is the module's own, not the
+        ``lock_timeout`` a caller configures — the two locks guard
+        different things.
+
+    .. rubric:: Implementation notes
+
+    `SharedMemory.__init__` registers the name *after* the block that
+    truncates, stats and maps it, but unlinks from inside that block's
+    ``except OSError`` handler. Below 3.13 `SharedMemory.unlink`
+    unregisters unconditionally; on 3.13 and up it unregisters whenever
+    ``track`` is set, which it is here. So on every supported version a
+    construction that fails mid-way unregisters a name that was never
+    registered — the tracker fault `_suppressing` describes, reached from
+    the owning side rather than the attaching one.
+
+    Only failures raised *inside* that block reach the handler. A
+    ``shm_open`` failure — ``EEXIST``, ``EMFILE``, ``ENOENT`` — raises
+    ahead of it and needs no suppression, which is what makes
+    `LocalDiscovery.__enter__`'s `FileExistsError` fallback safe.
+
+    Unlike the attach case, the ``shm_unlink`` that same handler performs
+    is correct here — the segment it destroys is the one this call just
+    created — so nothing is left for CPython to fix.
+
+    There is no version fast path. `_attach` can ask for ``track=False``
+    on 3.13 and skip the window entirely; nothing equivalent exists for a
+    create, so this rebinds the tracker hook on every supported version.
+    """
+    with _suppressing(name, register=False):
+        return SharedMemory(name=name, create=True, size=size)
+
+
 def _attach(name: str) -> SharedMemory:
     """Map an existing shared memory segment without tracking it.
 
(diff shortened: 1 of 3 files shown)


conradbzura:
Closing unmerged. The remedy for #340 has moved to #353, which backs the discovery registry with a memory-mapped file and takes `multiprocessing.resource_tracker` out of the picture entirely. That makes this fault unreachable by construction, and removes the `_suppressing`/`_create` machinery this PR builds out.

#340 stays open and will be closed by the PR that lands #353. Worth carrying forward from here: the page-boundary capacity property, the `os.ftruncate` failure injection (the file-backed registry still truncates before mapping), and running the lock liveness tests in a subprocess.

